Issue #132, opened 2014-04-10
Create XML Sitemap


ElectricMaxxx:
As we have access to the routes inside the PHPCR and the cmf handles routes it would be easy to create a xml representation of their structure.

But i would not only suggest to create a xml Stitemap, even a html representation would be doable. Maybe could add a configuration value for a template to render beautyfull lists of Routes, maybe with some previews to the content. 

The xml-generation could be done by `SonataSeoBundle`. The other thing need be done by us.

ElectricMaxxx:
solved by #196

PR #196, opened 2014-11-07
Sitemap generation

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #132 
| License       | MIT
| Doc PR        | https://github.com/symfony-cmf/symfony-cmf-docs/issues/607

Started working on that feature. That would be the task:

- [x] implement generator for phpcr content
- [x] create decider for content which should be published
- [x] create controller for the output
- [x] general configuration and optional activating of the phpcr generator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Changelog
 =========
 
+* **2015-14-02**: implement sitemap generation
+* **2015-14-02**: [BC BREAK] Changed method visibility from private to public of SeoPresentation#getSeoMetadata()
 * **2014-10-04**: Custom exception controller for error handling
 
 1.1.0-RC3
--- a/Controller/SitemapController.php
+++ b/Controller/SitemapController.php
@@ -1,0 +1,103 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2014 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Controller;
+
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+use Symfony\Cmf\Bundle\SeoBundle\Sitemap\UrlInformationProviderInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Templating\EngineInterface;
+
+/**
+ * Controller to handle requests for sitemap.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+class SitemapController
+{
+    /**
+     * @var UrlInformationProviderInterface
+     */
+    private $urlProvider;
+    /**
+     * @var EngineInterface
+     */
+    private $templating;
+
+    /**
+     * @var array
+     */
+    private $templates;
+
+    /**
+     * You should provide templates for html and xml.
+     *
+     * Json is serialized by default, but can be customized with a template
+     *
+     * @param UrlInformationProviderInterface $provider
+     * @param EngineInterface                 $templating
+     * @param array                           $templates  Hash map with key being the format,
+     *                                                    value the name of the twig template
+     *                                                    to render the sitemap in that format
+     */
+    public function __construct(
+        UrlInformationProviderInterface $provider,
+        EngineInterface $templating,
+        array $templates
+    ) {
+        $this->urlProvider = $provider;
+        $this->templating = $templating;
+        $this->templates = $templates;
+    }
+
+    /**
+     * @param string $_format The format of the sitemap.
+     *
+     * @return Response
+     */
+    public function indexAction($_format)
+    {
+        $supportedFormats = array_merge(array('json'), array_keys($this->templates));
+        if (!in_array($_format, $supportedFormats)) {
+            $text = sprintf(
+                'Unknown format %s, use one of %s.',
+                $_format,
+                implode(', ', $supportedFormats)
+            );
+
+            return new Response($text, 406);
+        }
+
+        $urls = $this->urlProvider->getUrlInformation();
+        if (isset($this->templates[$_format])) {
+            return new Response($this->templating->render($this->templates[$_format], array('urls' => $urls)));
+        }
+
+        return $this->createJsonResponse($urls);
+    }
+
+    /**
+     * @param array|UrlInformation[] $urls
+     *
+     * @return JsonResponse
+     */
+    private function createJsonResponse($urls)
+    {
+        $result = array();
+
+        foreach ($urls as $url) {
+            $result[] = $url->toArray();
+        }
+
+        return new JsonResponse($result);
+    }
+}
--- a/DependencyInjection/CmfSeoExtension.php
+++ b/DependencyInjection/CmfSeoExtension.php
@@ -14,6 +14,7 @@ namespace Symfony\Cmf\Bundle\SeoBundle\DependencyInjection;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Cmf\Bundle\RoutingBundle\Routing\DynamicRouter;
@@ -64,8 +65,6 @@ class CmfSeoExtension extends Extension
             $sonataBundles[] = 'SonataDoctrinePHPCRAdminBundle';
 
             $this->loadPhpcr($config['persistence']['phpcr'], $loader, $container);
-
-            $loader->load('matcher_phpcr.xml');
         }
 
         if ($this->isConfigEnabled($container, $config['persistence']['orm'])) {
@@ -87,6 +86,10 @@ class CmfSeoExtension extends Extension
 
         $errorConfig = isset($config['error']) ? $config['error'] : array();
         $this->loadErrorHandling($errorConfig, $container);
+
+        if ($this->isConfigEnabled($container, $config['sitemap'])) {
+            $this->loadSitemapHandling($config['sitemap'], $loader, $container);
+        }
     }
 
     /**
@@ -169,6 +172,9 @@ class CmfSeoExtension extends Extension
                 $this->defaultAlternateLocaleProviderId = 'cmf_seo.alternate_locale.provider_phpcr';
             }
         }
+
+        $loader->load('matcher_phpcr.xml');
+        $loader->load('phpcr-sitemap.xml');
     }
 
     /**
@@ -189,13 +195,19 @@ class CmfSeoExtension extends Extension
 
 
         if ($alternateLocaleProvider) {
-            $definition = $container->getDefinition('cmf_seo.event_listener.seo_content');
-            $definition
+            $alternateLocaleProviderDefinition = $container->findDefinition($alternateLocaleProvider);
+            $container
+                ->findDefinition('cmf_seo.event_listener.seo_content')
                 ->addMethodCall(
                     'setAlternateLocaleProvider',
-                    array($container->getDefinition($alternateLocaleProvider))
-                )
-            ;
+                    array($alternateLocaleProviderDefinition)
+                );
+            $container
+                ->findDefinition('cmf_seo.sitemap.phpcr_provider')
+                ->addMethodCall(
+                    'setAlternateLocaleProvider',
+                    array($alternateLocaleProviderDefinition)
+                );
         }
     }
 
@@ -210,10 +222,21 @@ class CmfSeoExtension extends Extension
     private function loadErrorHandling($config, ContainerBuilder $container)
     {
         foreach (array('parent', 'sibling') as $group) {
-            $remove = isset($config['enable_'.$group.'_provider']) && !$config['enable_'.$group.'_provider'] ? true : false;
+            $remove = isset($config['enable_'.$group.'_provider'])
+                    && !$config['enable_'.$group.'_provider'] ? true : false;
             if ($container->has('cmf_seo.error.suggestion_provider.'.$group) && $remove) {
                 $container->removeDefinition('cmf_seo.error.suggestion_provider.'.$group);
             }
         }
+    }
+
+    private function loadSitemapHandling($config, XmlFileLoader $loader, ContainerBuilder $container)
+    {
+        $loader->load('sitemap.xml');
+
+        $container->setParameter(
+            $this->getAlias().'.sitemap.default_change_frequency',
+            $config['default_chan_frequency']
+        );
     }
 }
--- a/DependencyInjection/Compiler/RegisterUrlInformationProviderPass.php
+++ b/DependencyInjection/Compiler/RegisterUrlInformationProviderPass.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2014 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\LogicException;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Compiler adds custom url information provider to the phpcr chain provider.
+ *
+ * To do so you need to tag them with "cmf_seo.sitemap.url_information_provider".
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+class RegisterUrlInformationProviderPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritDoc}
+     *
+     * @throws LogicException If a tagged service is not public.
+     */
+    public function process(ContainerBuilder $container)
+    {
+        // feature not activated means nothing to add
+        if (!$container->hasDefinition('cmf_seo.sitemap.url_information_provider')) {
+            return;
+        }
+
+        $chainProviderDefinition = $container->getDefinition('cmf_seo.sitemap.url_information_provider');
+        $taggedServices = $container->findTaggedServiceIds('cmf_seo.sitemap.url_information_provider');
+
+        foreach ($taggedServices as $id => $attributes) {
+            $priority = null;
+            foreach ($attributes as $attribute) {
+                if (isset($attribute['priority'])) {
+                    $priority = $attribute['priority'];
+                    break;
+                }
+            }
+            $priority = $priority ?: 0;
+
+            $chainProviderDefinition->addMethodCall('addProvider', array(new Reference($id), $priority));
+        }
+    }
+}
--- a/Doctrine/Phpcr/SitemapUrlInformationProvider.php
+++ b/Doctrine/Phpcr/SitemapUrlInformationProvider.php
@@ -1,0 +1,172 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2014 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr;
+
+use Doctrine\ODM\PHPCR\DocumentManager;
+use Doctrine\ODM\PHPCR\Query\QueryException;
+use Jackalope\Transport\Logging\LoggerInterface;
+use PHPCR\Query\QueryInterface;
+use Psr\Log\LoggerInterface as PsrLogger;
+use Symfony\Cmf\Bundle\CoreBundle\PublishWorkflow\PublishWorkflowChecker;
+use Symfony\Cmf\Bundle\SeoBundle\AlternateLocaleProviderInterface;
+use Symfony\Cmf\Bundle\SeoBundle\Extractor\ExtractorInterface;
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+use Symfony\Cmf\Bundle\SeoBundle\SeoPresentation;
+use Symfony\Cmf\Bundle\SeoBundle\Sitemap\UrlInformationProviderInterface;
+use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Core\SecurityContextInterface;
+
+/**
+ * Provides UrlInformation for pages in a Doctrine PHPCR ODM back-end.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+class SitemapUrlInformationProvider implements UrlInformationProviderInterface
+{
+    /**
+     * @var AlternateLocaleProviderInterface
+     */
+    protected $alternateLocaleProvider;
+
+    /**
+     * @var ExtractorInterface
+     */
+    protected $titleExtractor;
+
+    /**
+     * @var DocumentManager
+     */
+    private $manager;
+
+    /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    /**
+     * @var SecurityContextInterface
+     */
+    private $publishWorkflowChecker;
+    /**
+     * @var string
+     */
+    private $defaultChanFrequency;
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+    /**
+     * @var SeoPresentation
+     */
+    private $seoPresentation;
+
+    /**
+     * @param DocumentManager $manager
+     * @param RouterInterface $router
+     * @param string $defaultChanFrequency
+     * @param PsrLogger $logger
+     * @param SeoPresentation $seoPresentation
+     * @param SecurityContextInterface $publishWorkflowChecker
+     */
+    public function __construct(
+        DocumentManager $manager,
+        RouterInterface $router,
+        $defaultChanFrequency,
+        PsrLogger $logger,
+        SeoPresentation $seoPresentation,
+        SecurityContextInterface $publishWorkflowChecker = null
+    ) {
+        $this->manager = $manager;
+        $this->router = $router;
+        $this->defaultChanFrequency = $defaultChanFrequency;
+        $this->logger = $logger;
+        $this->seoPresentation = $seoPresentation;
+        $this->publishWorkflowChecker = $publishWorkflowChecker;
+    }
+
+    /**
+     * {@inheritDocs}
+     */
+    public function getUrlInformation()
+    {
+        $contentDocuments = $this->fetchSitemapDocuments();
+        $urlInformationList = array();
+
+        foreach ($contentDocuments as $document) {
+            if (null != $this->publishWorkflowChecker &&
+                !$this->publishWorkflowChecker->isGranted(array(PublishWorkflowChecker::VIEW_ATTRIBUTE), $document)
+            ) {
+                continue;
+            }
+
+            try {
+                $urlInformationList[] = $this->computeUrlInformationFromSitemapDocument($document);
+            } catch (\Exception $e) {
+                $this->logger->info($e->getMessage());
+            }
+        }
+
+        return $urlInformationList;
+    }
+
+    /**
+     * @param AlternateLocaleProviderInterface $alternateLocaleProvider
+     */
+    public function setAlternateLocaleProvider(AlternateLocaleProviderInterface $alternateLocaleProvider)
+    {
+        $this->alternateLocaleProvider = $alternateLocaleProvider;
+    }
+
+    /**
+     * Wrapper to fetch the sitemap documents from database.
+     *
+     * @return object[]
+     *
+     * @throws QueryException
+     */
+    protected function fetchSitemapDocuments()
+    {
+        // todo rewrite query when https://github.com/symfony-cmf/CoreBundle/issues/126 is ready
+        return $this->manager->createQuery(
+            "SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true)",
+            QueryInterface::JCR_SQL2
+        )->execute();
+    }
+
+    /**
+     * Transforms a single sitemap document into url information.
+     *
+     * A sitemap document is a document, which should be exposed on a sitemap.
+     *
+     * @param object $document
+     *
+     * @return UrlInformation
+     */
+    protected function computeUrlInformationFromSitemapDocument($document)
+    {
+        $urlInformation = new UrlInformation();
+        $urlInformation->setLocation($this->router->generate($document, array(), true));
+        $urlInformation->setChangeFrequency($this->defaultChanFrequency);
+
+        if ($this->alternateLocaleProvider) {
+            $collection = $this->alternateLocaleProvider->createForContent($document);
+            $urlInformation->setAlternateLocales($collection->toArray());
+        }
+
+        $seoMetadata = $this->seoPresentation->getSeoMetadata($document);
+        if (null !== $seoMetadata->getTitle()) {
+            $urlInformation->setLabel($seoMetadata->getTitle());
+            return $urlInformation;
+        }
+        return $urlInformation;
+    }
+}
--- a/Model/UrlInformation.php
+++ b/Model/UrlInformation.php
@@ -1,0 +1,231 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2014 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Model;
+use Symfony\Cmf\Bundle\SeoBundle\Exception\InvalidArgumentException;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+class UrlInformation
+{
+    /**
+     * Decides whether the content is visible on sitemap.
+     *
+     * @var bool
+     */
+    private $visible;
+
+    /**
+     * @var string
+     */
+    private $location;
+
+    /**
+     * @var \DateTime
+     */
+    private $lastModification;
+
+    /**
+     * @var string One of the official/allowed.
+     */
+    private $changeFrequency;
+
+    /**
+     * @var float
+     */
+    private $priority;
+
+    /**
+     * @var array
+     */
+    private $allowedChangeFrequencies = array('always', 'hourly', 'daily', 'weekly', 'monthly', 'yearly', 'never');
+
+    /**
+     * @var string $label As a string to display the route i.e. in html views.
+     */
+    private $label;
+
+    /**
+     * @var AlternateLocale[]
+     */
+    private $alternateLocales;
+
+    public function __construct()
+    {
+        $this->alternateLocales = array();
+    }
+
+    /**
+     * @return boolean
+     */
+    public function isVisible()
+    {
+        return $this->visible;
+    }
+
+    /**
+     * @param boolean $visible
+     */
+    public function setVisible($visible)
+    {
+        $this->visible = $visible;
+    }
+
+    public function toArray()
+    {
+        $result = array(
+            'loc'               => $this->location,
+            'label'             => $this->label,
+            'changefreq'        => $this->changeFrequency,
+            'lastmod'           => $this->lastModification,
+            'priority'          => $this->priority,
+            'alternate_locales' => array()
+        );
+
+        foreach ($this->alternateLocales as $locale) {
+            $result['alternate_locales'][] = array('href' => $locale->href, 'href_locale' => $locale->hrefLocale);
+        }
+
+        return $result;
+    }
+
+    /**
+     * @return string
+     */
+    public function getChangeFrequency()
+    {
+        return $this->changeFrequency;
+    }
+
+    /**
+     * @param string $changeFrequency One of the official/allowed ones.
+     *
+     * @return $this
+     */
+    public function setChangeFrequency($changeFrequency)
+    {
+        if (!in_array($changeFrequency, $this->allowedChangeFrequencies)) {
+            throw new InvalidArgumentException(
+                sprintf('Invalid change frequency use one of %s.', implode(', ', $this->allowedChangeFrequencies))
+            );
+        }
+
+        $this->changeFrequency = $changeFrequency;
+
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getLastModification()
+    {
+        return $this->lastModification;
+    }
+
+    /**
+     * @param \DateTime $dateTime
+     *
+     * @return $this
+     */
+    public function setLastModification(\DateTime $dateTime)
+    {
+        $lastmod = $dateTime->format('c');
+        $this->lastModification = $lastmod;
+
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getLocation()
+    {
+        return $this->location;
+    }
+
+    /**
+     * @param string $location
+     *
+     * @return $this
+     */
+    public function setLocation($location)
+    {
+        $this->location = $location;
+
+        return $this;
+    }
+
+    /**
+     * @return float
+     */
+    public function getPriority()
+    {
+        return $this->priority;
+    }
+
+    /**
+     * @param float $priority
+     *
+     * @return $this
+     */
+    public function setPriority($priority)
+    {
+        $this->priority = $priority;
+
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getLabel()
+    {
+        return $this->label;
+    }
+
+    /**
+     * @param string $label
+     *
+     * @return $this
+     */
+    public function setLabel($label)
+    {
+        $this->label = $label;
+
+        return $this;
+    }
+
+    /**
+     * @return array|AlternateLocale[]
+     */
+    public function getAlternateLocales()
+    {
+        return $this->alternateLocales;
+    }
+
+    /**
+     * @param array|AlternateLocale[] $alternateLocales
+     */
+    public function setAlternateLocales($alternateLocales)
+    {
+        $this->alternateLocales = $alternateLocales;
+    }
+
+    /**
+     * @param AlternateLocale $alternateLocale
+     */
+    public function addAlternateLocale(AlternateLocale $alternateLocale)
+    {
+        $this->alternateLocales[] = $alternateLocale;
+    }
+}
--- a/Resources/config/phpcr-sitemap.xml
+++ b/Resources/config/phpcr-sitemap.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" ?>
+
+<container
+        xmlns="http://symfony.com/schema/dic/services"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <parameters>
+        <parameter key="cmf_seo.sitemap.phpcr_provider.class">Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr\SitemapUrlInformationProvider</parameter>
+    </parameters>
+
+    <services>
+        <service id="cmf_seo.sitemap.phpcr_provider" class="%cmf_seo.sitemap.phpcr_provider.class%" public="false">
+            <argument type="service" id="doctrine_phpcr.odm.default_document_manager"/>
+            <argument type="service" id="router" />
+            <argument>%cmf_seo.sitemap.default_change_frequency%</argument>
+            <argument type="service" id="logger" />
+            <argument type="service" id="cmf_seo.presentation" />
+            <argument type="service" id="cmf_core.publish_workflow.checker" />
+
+            <tag name="cmf_seo.sitemap.url_information_provider" />
+        </service>
+    </services>
+</container>
--- a/Resources/config/routing/sitemap.xml
+++ b/Resources/config/routing/sitemap.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<routes xmlns="http://symfony.com/schema/routing"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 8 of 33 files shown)


Resources/config/routing/sitemap.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<routes xmlns="http://symfony.com/schema/routing"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/routing http://symfony.com/schema/routing/routing-1.0.xsd">
+
+    <route id="cmf_seo_sitemap" pattern="/sitemap.{_format}">
+        <default key="_controller">cmf_seo.sitemap.controller:indexAction</default>
+        <default key="_format">html</default>
```
ElectricMaxxx: done.

ElectricMaxxx:
Except the decission how to get the tilte (seo extractor, method call or some own extractor?) everything should be solved.

Controller/SitemapController.php
```diff
@@ -0,0 +1,107 @@
...
+ * Controller to handle requests for sitemap.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+class SitemapController
+{
+    /**
+     * @var UrlInformationProviderInterface
+*/
+    private $chainProvider;
```
dbu: this should be called urlProvider, its not necessarily the chain provider.
and indention of comment close above is wrong.

Controller/SitemapController.php
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * @var array
+     */
+    private $templates;
+
+    /**
+     * Inject the templates as an hashmap like:
+     *
+     * array('format' => 'Bundle:Domain:template.format.twig')
```
dbu: lets write this as '<format>' and say that the valid formats are html, json and xml.

Controller/SitemapController.php
```diff
@@ -0,0 +1,107 @@
...
+    public function indexAction($_format)
+    {
+        $response = null;
+        $urls = $this->chainProvider->generateRoutes();
+
+        if ('json' === $_format) {
+            $response = $this->createJsonResponse($urls);
+        } elseif (isset($this->templates[$_format])) {
+            $response =  new Response($this->templating->render($this->templates[$_format], array('urls' => $urls)));
+        }
```
dbu: maybe we could also allow other formats and remove the format requirement from the route. then you could start with isset (so that i can specify a template for json too), then try json and otherwise throw a 404 that this page is not found. with a message which formats are supported as you do below.

DependencyInjection/Compiler/RegisterUrlInformationProviderPass.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Symfony\Cmf\Bundle\SeoBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\LogicException;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
```
dbu: please add one line what this does (register services tagged with ... to the sitemap chain provider.)

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,129 @@
...
+        $urlInformation->setLoc($this->router->generate($content, array(), true));
+        $urlInformation->setChangeFreq($this->defaultChanFrequency);
+
+        if ($this->alternateLocaleProvider) {
+            $collection = $this->alternateLocaleProvider->createForContent($content);
+            $urlInformation->setAlternateLocales($collection->toArray());
+        }
+
+        if (method_exists($content, 'getTitle')) {
+            $urlInformation->setLabel($content->getTitle());
```
dbu: did you look into that? i am unhappy with duplicating the title extractor logic in an inferior way here.

dbu: if we go through SeoMetadata, that seems ok to me i think. a bit heavy, but it will be heavy anyways.

SeoPresentation.php
```diff
@@ -136,14 +136,9 @@ public function addExtractor(ExtractorInterface $extractor, $priority = 0)
...
-     * Gets the SeoMetadata based on the content that contains the content.
-     *
-     * @param object $content
-     *
-     * @throws Exception\InvalidArgumentException
-     * @return SeoMetadata
+     * {@inheritDoc}
      */
-    private function getSeoMetadata($content)
+    public function getSeoMetadata($content)
```
ElectricMaxxx: I am not really sure if that is a BC-Break? Otherwise i would implement a little wrapper method, which calls that. But calling that method would use the complete seo metadata creation including the caching for the extractors.

dbu: no, relaxing access to a method never is a BC break.

SeoPresentationInterface.php
```diff
@@ -23,6 +24,16 @@
...
 {
     /**
+     * Gets the SeoMetadata based on the content that contains the content.
+     *
+     * @param object $content
+     *
+     * @throws Exception\InvalidArgumentException
+     * @return SeoMetadata
+     */
+    public function getSeoMetadata($content);
```
dbu: this however is a BC break

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,160 @@
...
+     * @param RouterInterface $router
+     * @param SecurityContextInterface $publishWorkflowChecker
+     * @param string $defaultChanFrequency
+     * @param PsrLogger $logger
+     * @param SeoPresentationInterface $seoPresentation
+     */
+    public function __construct(
+        DocumentManager $manager,
+        RouterInterface $router,
+        SecurityContextInterface $publishWorkflowChecker,
```
dbu: the publish workflow checker is optional, we should allow this to be `null` and skip published check if its not set.

SeoPresentationInterface.php
dbu: maybe we should just have the phpcr sitemap provider depend on the concrete implementation instead of adding this to the interface.

SitemapElementInterface.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Symfony\Cmf\Bundle\SeoBundle;
+
+/**
+ * Documents persisted with phpcr-odm should implement that interface.
+ * They also need to create a property with the name "visible_for_sitemap" to be recognized
+ * by the default provider, that generates the UrlInformation object.
```
dbu: lets say:

Models implementing this interface can indicate whether they should be listed in the sitemap or not.

For PHPCR-ODM documents used with the Doctrine\Phpcr\Sitemap.. , make sure to call the property that stores this information "visible_for_sitemap", as the provider uses the property in the query rather than loading all documents and checking the method.

SitemapElementInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+interface SitemapElementInterface
```
dbu: one potential problem i see with this: there could be more than one sitemap, e.g. an overview for human visitors and the xml sitemap to help google. visibility could depend on the type of sitemap...

i don't want to solve this problem in this PR, but maybe we can quickly think about it to avoid having a solution that blocks further flexibility.

Sitemap/UrlInformationProviderInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Sitemap;
+
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+
+/**
+ * Interface for a provider that is able to create a list of
+ * UrlInformation for sitemap creation.
```
dbu: Providers are able to create a list of UrlInformation for the sitemap creation.

Sitemap/UrlInformationProviderInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+interface UrlInformationProviderInterface
+{
+    /**
+     * @return UrlInformation[]
+     */
+    public function generateRoutes();
```
dbu: what about "getUrlInformation"?

and we should describe the method a little bit: "Provides an array of UrlInformation to build the sitemap."

SitemapElementInterface.php
dbu: maybe we could drop this interface as we do not use it anyways, and could instead configure the phpcr-odm provider with a list of properties of which at least one must be true for a document to be included. if no properties are configured, there is no filtering and all routable documents end up in the sitemap.

dbu: then i could do visible_html_sitemap and visible_google_xml_sitemap for example.

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Sitemap;
+
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+class ChainProvider implements UrlInformationProviderInterface
```
dbu: we should describe this class: "Accepts providers and merges the result of all providers into a combined list of UrlInformation"

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    /**
+     * @var UrlInformationProviderInterface[]
+     */
+    private $providers = array();
+
+    /**
+     * @return UrlInformationProviderInterface[]
+     */
+    public function getProviders()
```
dbu: this method seems unused. lets remove it.

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        return $this->providers;
+    }
+
+    /**
+     * @param UrlInformationProviderInterface[] $providers
+     */
+    public function setProviders($providers)
```
dbu: this method is unused, lets remove it

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $this->providers = $providers;
+    }
+
+    public function addProvider(UrlInformationProviderInterface $provider)
```
dbu: does order matter? if so, we should have an optional priority parameter, and keep a private getProviders that sorts the providers by priority. (see https://github.com/symfony-cmf/Routing/blob/master/ChainRouter.php for example)

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $this->providers[] = $provider;
+    }
+
+    /**
+     * @return UrlInformation[]
```
dbu: `{@inheritDoc}`

Combines the output of all registered providers into one array.

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,160 @@
...
+    {
+        $routeInformation = array();
+
+        $contentDocuments = $this->manager->createQuery(
+            "SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true)",
+            QueryInterface::JCR_SQL2
+        )->execute();
+
+        foreach ($contentDocuments as $document) {
+            if (!$this->publishWorkflowChecker->isGranted(array(PublishWorkflowChecker::VIEW_ATTRIBUTE), $document)) {
```
dbu: we really should do https://github.com/symfony-cmf/CoreBundle/issues/126 at some point - then you could ask the publish workflow checker to add this to the query.

Controller/SitemapController.php
```diff
@@ -0,0 +1,108 @@
...
+        } elseif ('json' === $_format) {
+            $response = $this->createJsonResponse($urls);
+        }
+
+        if (null === $response) {
+            $supportedFormats = array_keys($this->templates);
+            $supportedFormats[] = 'json';
+            throw new InvalidConfigurationException(
+                sprintf(
+                    'Type %s is not configures for sitemaps, use one of %s',
```
dbu: this should be a HTTP 406 error, or maybe 404.

i would make the message: "Unknown format %s, use one of %s."

Controller/SitemapController.php
```diff
@@ -0,0 +1,108 @@
...
+        $urls = $this->urlProvider->generateRoutes();
+        if (isset($this->templates[$_format])) {
+            $response =  new Response($this->templating->render($this->templates[$_format], array('urls' => $urls)));
+        } elseif ('json' === $_format) {
+            $response = $this->createJsonResponse($urls);
+        }
+
+        if (null === $response) {
+            $supportedFormats = array_keys($this->templates);
+            $supportedFormats[] = 'json';
```
dbu: if (!in_array('json', $supportedFormats[])) { ...

Controller/SitemapController.php
```diff
@@ -0,0 +1,108 @@
...
+        EngineInterface $templating,
+        array $templates
+    ) {
+        $this->urlProvider = $provider;
+        $this->templating = $templating;
+        $this->templates = $templates;
+    }
+
+    /**
+     * @param string $_format The format of the sitemap. Supported values are html|xml|json
```
dbu: Remove the thing about supported values, now that we allow to inject any templates.

Controller/SitemapController.php
```diff
@@ -0,0 +1,108 @@
...
+     * @var array
+     */
+    private $templates;
+
+    /**
+     * Inject the templates as an hashmap like:
+     *
+     * array('<format>' => 'Bundle:Domain:template.format.twig')
+     *
+     * Valid formats are html, json and xml.
```
dbu: i would say: "You should provide templates for html and xml. Json is serialized by default, but can be customized with a template."

dbu:
alright, i think we really are getting there!

i went over details and proposed some simplifications.

SitemapElementInterface.php
ElectricMaxxx: What would be the decission in here?
Droppping the interface and implement an multi sitemap sollution later?

Doctrine/Phpcr/SitemapUrlInformationProvider.php
ElectricMaxxx: I would suggest to add a todo here.

ElectricMaxxx: and https://github.com/symfony-cmf/SeoBundle/issues/210

SeoPresentationInterface.php
ElectricMaxxx: @dbu That means? Remove the method from the interface and ...?

ElectricMaxxx:
Rebased and i think i solved all issues mentioned in the comments. I would squash it if you like it.

SeoPresentationInterface.php
dbu: remove the method from the interface and meke the phpcr provider typehint the concrete class rather than the interface. it is bound to the implementation anyways.

Controller/SitemapController.php
```diff
@@ -0,0 +1,101 @@
...
+    private $templates;
+
+    /**
+     * You should provide templates for html and xml.
+     *
+     * Json is serialized by default, but can be customized with a template
+     *
+     * @param UrlInformationProviderInterface $provider
+     * @param EngineInterface $templating
+     * @param array $templates
```
dbu: can we say: "Hash map with key being the format, value the name of the twig template to render the sitemap in that format"?

Resources/config/phpcr-sitemap.xml
```diff
@@ -0,0 +1,24 @@
...
+        xmlns="http://symfony.com/schema/dic/services"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <parameters>
+        <parameter key="cmf_seo.sitemap.phpcr_provider.class">Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr\SitemapUrlInformationProvider</parameter>
+    </parameters>
+
+    <services>
+        <service id="cmf_seo.sitemap.phpcr_provider" class="%cmf_seo.sitemap.phpcr_provider.class%">
```
dbu: lets make this service private. it is always loaded even when not needed. and there is no point getting it from the container.

dbu: a private service can still be used in service definitions and compiler passes.

DependencyInjection/Compiler/RegisterUrlInformationProviderPass.php
```diff
@@ -0,0 +1,52 @@
...
+            return;
+        }
+
+        $chainProviderDefinition = $container->getDefinition('cmf_seo.sitemap.chain_provider');
+        $taggedServices = $container->findTaggedServiceIds('cmf_seo.sitemap.url_information_provider');
+
+        foreach ($taggedServices as $id => $attributes) {
+            $definition = $container->getDefinition($id);
+            if (!$definition->isPublic()) {
+                throw new LogicException(sprintf('Provider "%s" must be public.', $id));
```
dbu: i think this is not correct. the service can be private.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -66,6 +66,7 @@ public function load(array $configs, ContainerBuilder $container)
             $this->loadPhpcr($config['persistence']['phpcr'], $loader, $container);
 
             $loader->load('matcher_phpcr.xml');
+            $loader->load('phpcr-sitemap.xml');
```
dbu: but maybe both this line and the line above could go to into $this->loadPhpcr to keep load() shorter

DependencyInjection/CmfSeoExtension.php
```diff
@@ -216,4 +230,16 @@ private function loadErrorHandling($config, ContainerBuilder $container)
             }
         }
     }
+
+    private function loadSitemapHandling($config, XmlFileLoader $loader, ContainerBuilder $container)
+    {
+        if ($config['enabled']) {
```
dbu: i think you don't need the check here, its already checked to see if loadSitemapHandling needs to be called.

Controller/SitemapController.php
ElectricMaxxx: as description for the `template` paramter section?

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,163 @@
...
+     *
+     * @param $content
+     *
+     * @return UrlInformation
+     */
+    protected function computeUrlInformationFromContent($content)
+    {
+        $urlInformation = new UrlInformation();
+        $urlInformation->setLoc($this->router->generate($content, array(), true));
+        $urlInformation->setChangeFreq($this->defaultChanFrequency);
```
dbu: i guess we could have an extractor for that as well, but we should push that to a next version.

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * @param AlternateLocaleProviderInterface $alternateLocaleProvider
+     */
+    public function setAlternateLocaleProvider($alternateLocaleProvider)
+    {
+        $this->alternateLocaleProvider = $alternateLocaleProvider;
+    }
+
+    public function setTitleExtractor(ExtractorInterface $titleExtractor)
```
dbu: now that we use the seoPresentation, this is not used ever, right?

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,163 @@
...
+            $urlInformation->setLabel($seoMetadata->getTitle());
+        }
+
+        return $urlInformation;
+    }
+
+    /**
+     * @param AlternateLocaleProviderInterface $alternateLocaleProvider
+     */
+    public function setAlternateLocaleProvider($alternateLocaleProvider)
```
dbu: why is this not an optional dependency like the publish workflow checker?

DependencyInjection/Compiler/RegisterUrlInformationProviderPass.php
ElectricMaxxx: just a copy of other tag fetching compilers :-)
just realized it in the tests

SitemapElementInterface.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Symfony\Cmf\Bundle\SeoBundle;
+
+/**
+ * Models implementing this interface can indicate whether they should be listed in the sitemap or not.
+ *
+ * For PHPCR-ODM documents used with the Doctrine\Phpcr\Sitemap.. , make sure to call the property that stores
```
dbu: "the Doctrine\Phpcr\Sitemap.." - can we give the full bundle-relative class name ehre?

dbu:
sorry, i just re-reviewed all things and found a couple of minor things. would really love to wrap this up - but do you think you can check these details?

one question that just popped to my mind: could we optionall tell the phpcr provider to just expose all documents that have a route and are published, rather than rely on the interface/property stored in phpcr?

DependencyInjection/CmfSeoExtension.php
ElectricMaxxx:
``` 
if ($this->isConfigEnabled($container, $config['sitemap'])) {
...
```

is enough, right?

Controller/SitemapController.php
dbu: yep, exactly.

Doctrine/Phpcr/SitemapUrlInformationProvider.php
ElectricMaxxx: thank's :-) i would see it as a kind of refactoring.

DependencyInjection/Compiler/RegisterUrlInformationProviderPass.php
dbu: okay. i think we have some that do not have this mistake

Doctrine/Phpcr/SitemapUrlInformationProvider.php
ElectricMaxxx: :+1:

DependencyInjection/CmfSeoExtension.php
dbu: yeah, i would hope so.

Doctrine/Phpcr/SitemapUrlInformationProvider.php
dbu: right, created #212 - no need to do this right now

Doctrine/Phpcr/SitemapUrlInformationProvider.php
ElectricMaxxx: cause its available on alternate locale activation only

dbu: but you could just alter the service definition to add one more constructor argument then, instead of adding a method call. but then its not super important either, i am ok with keeping as is.

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,163 @@
...
+
+        $seoMetadata = $this->seoPresentation->getSeoMetadata($content);
+        if (null !== $seoMetadata->getTitle()) {
+            $urlInformation->setLabel($seoMetadata->getTitle());
+        }
+
+        return $urlInformation;
+    }
+
+    /**
```
dbu: can you say that this is optional and when set makes the sitemap provide alternate locale urls?

Doctrine/Phpcr/SitemapUrlInformationProvider.php
dbu: oh, but you really should add a typehint to the method.

Doctrine/Phpcr/SitemapUrlInformationProvider.php
ElectricMaxxx: added some text in the method documentation

ElectricMaxxx:
so ... fixed your comments on the fly. Now i need to understand your question.

ElectricMaxxx:
@dbu you mean to devide the provider into two parts, one that fetches the routes from the database (to reuse it somewhere else) and a kind of an "MakeUrlInformationFromRoutesService"?

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,161 @@
...
+     */
+    public function getUrlInformation()
+    {
+        $routeInformation = array();
+
+        // todo rewrite query when https://github.com/symfony-cmf/CoreBundle/issues/126 is ready
+        $contentDocuments = $this->manager->createQuery(
+            "SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true)",
+            QueryInterface::JCR_SQL2
+        )->execute();
```
dbu: maybe we can just move this bit to a protected `fetchSitemapDocuments` method. then it will be very easy to extend the class and adjust the query to other needs.

in the long run we could have two separate classes, one to fetch and one to transform. transforming an orm entity would probably look exactly the same, so it does not really belong in here. but these changes can be made without BC breaks, its just refactoring and moving things out.

i would suggest we just move creating and executing the query into a separate protected method.

ElectricMaxxx: but i would see many use cases to fetch the result before creating the url information from it. As you said it: "give all documents which are publish and have routes" i would expect that in repository alike service close to our core-bundle.

dbu: yeah, if the `$this->manager ... ->execute();` is in a protected method, at least people can extend this class to handle their use case. that would be good enough for the first version of this.

ElectricMaxxx:
144 comments ... i think i got it now.

DependencyInjection/CmfSeoExtension.php
dbu: can you move these into the loadPhpcr method?

dbu:
epic! but its an important and very cool feature! i will soon have to update some projects to start using this.

i have one more comment open, about moving loading phpcr config files into loadPhpcr. i think then i am happy. can you squash the commits then?

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /**
+     * {@inheritDocs}
+     */
+    public function getUrlInformation()
+    {
+        $contentDocuments = $this->fetchSitemapDocuments();
+
+        return $this->computeUrlInformationFromSitemapDocuments($contentDocuments);
```
dbu: i would suggest to keep the loop and published check in this method, and have the compute only handle a single document per call. makes it easier if i want to customize the compute

ElectricMaxxx:
@dbu no problems with the count of your hints. I am trying to inject each of them in my skills and getting better and better every day. You are a kind of a mentor for me.

dbu:
@ElectricMaxxx thanks, glad you appreciate it. i also learn a lot doing this. 

i don't find anything to change anymore. do you want to squash the commits or shall i?

ElectricMaxxx:
missed one, but got it now.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -169,6 +171,11 @@ private function loadPhpcr($config, XmlFileLoader $loader, ContainerBuilder $con
                 $this->defaultAlternateLocaleProviderId = 'cmf_seo.alternate_locale.provider_phpcr';
             }
         }
+
+
+
```
WouterJ: some redundant empty lines here

DependencyInjection/CmfSeoExtension.php
```diff
@@ -196,6 +203,15 @@ private function loadAlternateLocaleProvider($config, ContainerBuilder $containe
                     array($container->getDefinition($alternateLocaleProvider))
                 )
             ;
+
+            if ($container->has('cmf_seo.sitemap.phpcr_provider')) {
```
WouterJ: you should use `findDefinition`

DependencyInjection/CmfSeoExtension.php
```diff
@@ -196,6 +203,15 @@ private function loadAlternateLocaleProvider($config, ContainerBuilder $containe
...
+                $definition = $container->getDefinition('cmf_seo.sitemap.phpcr_provider');
```
WouterJ: same here

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * @var ExtractorInterface
+     */
+    protected $titleExtractor;
+
+    /**
+     * @var DocumentManager
+     */
+    private $manager;
+    /**
```
WouterJ: missing empty line before this one

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,169 @@
...
+use Symfony\Cmf\Bundle\SeoBundle\AlternateLocaleProviderInterface;
+use Symfony\Cmf\Bundle\SeoBundle\Extractor\ExtractorInterface;
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+use Symfony\Cmf\Bundle\SeoBundle\SeoPresentation;
+use Symfony\Cmf\Bundle\SeoBundle\Sitemap\UrlInformationProviderInterface;
+use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Core\SecurityContextInterface;
+
+/**
+ * Providers are able to create a list of UrlInformation for the sitemap creation.
```
WouterJ: let's tell the purpose of this provider instead of the purpose of providers in general, e.g: "Provides UrlInformation for pages in a Doctrine PHPCR ODM back-end."

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,169 @@
...
+     */
+    public function setAlternateLocaleProvider(AlternateLocaleProviderInterface $alternateLocaleProvider)
+    {
+        $this->alternateLocaleProvider = $alternateLocaleProvider;
+    }
+
+    /**
+     * Wrapper to fetch the sitemap documents from database.
+     *
+     * @return array []
```
WouterJ: you mean, array of arrays? In that case: `array[]`. Currently, the type is `array` and the description is `[]`

dbu: i think this should be `@return object[] List of PHPCR-ODM documents for the sitemap`

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,169 @@
...
+     *
+     * @throws \Doctrine\ODM\PHPCR\Query\QueryException
```
dbu: lets `use` this and then have the short name

Model/UrlInformation.php
```diff
@@ -0,0 +1,228 @@
...
+     * Decides whether the content is visible on sitemap.
+     *
+     * @var bool
+     */
+    private $visible;
+
+    /**
+     * @var string
+     */
+    private $loc;
```
WouterJ: I would love to use full names: `location` (and `lastModification`, `changeFrequency`, `allowedChangeFrequencies` belong)

Doctrine/Phpcr/SitemapUrlInformationProvider.php
```diff
@@ -0,0 +1,169 @@
...
+            QueryInterface::JCR_SQL2
+        )->execute();
+    }
+
+    /**
+     * Transforms a single sitemap document into url information.
+     *
+     * A sitemap document is a document, which should be exposed on a sitemap.
+     *
+     * @param $document
```
dbu: `@param object $document The content document to build the url information for.`.

Model/UrlInformation.php
```diff
@@ -0,0 +1,228 @@
...
+     */
+    private $allowedChangeFreqs = array('always', 'hourly', 'daily', 'weekly', 'monthly', 'yearly', 'never');
+
+    /**
+     * @var string $label As a string to display the route i.e. in html views.
+     */
+    private $label;
+
+    /**
+     * @var array|AlternateLocale[]
```
WouterJ: is it either an array with random types or an array with `AlternateLocale` instances, or is it always an array with `AlternateLocale` instances?

Model/UrlInformation.php
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * @param string $changeFreq One of the official/allowed ones.
+     *
+     * @return $this
+     */
+    public function setChangeFreq($changeFreq)
+    {
+        if (!in_array($changeFreq, $this->allowedChangeFreqs)) {
+            return $this;
```
WouterJ: Silently ignoring things makes debugging very hard. We should throw an exception here imo

Model/UrlInformation.php
```diff
@@ -0,0 +1,228 @@
...
+     */
+    public function isVisible()
+    {
+        return $this->visible;
+    }
+
+    /**
+     * @param boolean $visible
+     */
+    public function setVisible($visible)
```
WouterJ: we should ask ourselves if we really need setters for each property.

Resources/views/Sitemap/index.html.twig
```diff
@@ -0,0 +1,7 @@
+<ul class="cmf-sitemap">
+    {% for url in urls %}
+        <li>
+            <a href="{{ url.loc }}" title="{{ url.label }}">{{ url.label }}</a>
+        </li>
+    {% endfor %}
+</ul>
```
WouterJ: missing line feed character

SeoPresentation.php
WouterJ: It is when going from private to public, as a class overriding this method cannot make a method less visible (only more visible).

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Accepts providers and merges the result of all providers into
+ * a combined list of UrlInformation
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+class ChainProvider implements UrlInformationProviderInterface
+{
+    /**
+     * @var UrlInformationProviderInterface[][]
```
WouterJ: is this valid PHPdoc?

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * @return UrlInformation[]
+     */
+    public function getUrlInformation()
+    {
+        $urlInformation = array();
+
+        foreach ($this->getSortedProviders() as $provider) {
```
WouterJ: I would cache the sorted list and use it, to avoid sorting the list again and again

Model/UrlInformation.php
dbu: yeah, i was wondering the same. just a constructor could be very easy. but then, we might have various voters or extractors in some situations, where it becomes really cumbersome to re-create the UrlInformation constantly just to update some data.

all in all, this could be an array with keys - its really just a value container. so imho setters are fine in this case.

Tests/Functional/Doctrine/Phpcr/SitemapUrlInformationProviderTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * @var SitemapUrlInformationProvider
+     */
+    protected $provider;
+
+    protected $alternateLocaleProvider;
+
+    protected $logger;
+
```
WouterJ: no empty line required if you don't use phpdoc

Model/UrlInformation.php
dbu: note that i prefer the value object over an array, its explicit and documented over random array keys nobody knows. one of the really bad points of drupal, at least until 7.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* **2014-11-07**: implement sitemap generation
```
WouterJ: this date isn't correct anymore

Controller/SitemapController.php
```diff
@@ -0,0 +1,103 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2014 Symfony CMF
```
WouterJ: it's 2015 now :) (and as this file was added in 2015, it should officialy say "(c) 2015 Symfony CMF")

WouterJ:
I'm sorry to join this party so late, but I'm finished now. :)

SeoPresentation.php
dbu: actually not when the method was private afaik. not in php, that is. you can't have another private method in an extending class with the same name, nor a protected/public one, as its declared private here. even if i am wrong, i think this change is minor enough that i can live with it - SeoPresentation needs not be extended, there are the extractors for customizign.

afaik Java handles that with less WTF on language level (but i guess a lot of WTF when actually running into that situation).

Sitemap/ChainProvider.php
dbu: afaik not. but anyway it should be just one [] here

Sitemap/ChainProvider.php
dbu: that should happen in getSortedProviders, though

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public function addProvider(UrlInformationProviderInterface $provider, $priority = 0)
+    {
+        if (empty($this->providers[$priority])) {
+            $this->providers[$priority] = array();
+        }
+        $this->providers[$priority][] = $provider;
```
dbu: add: `$this->sortedProviders = array();`

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * @return UrlInformationProviderInterface[]
+     */
+    private function getSortedProviders()
+    {
+        $sortedProviders = array();
```
dbu: make this 

```
if (!$this->sortedProviders) {
... build the sorted list ...
}

return $this->sortedProviders;
```

SeoPresentation.php
WouterJ: @dbu http://3v4l.org/ECNUM PHP does bother (not the other way around though).

And yeah, I can live with it too, but it has to be added to the CHANGELOG with a `[BC BREAK]` prefix imo

Sitemap/ChainProvider.php
WouterJ: it should be 2, as the array is:

````
array(
    '<PRIORITY>' => array(
        ... instances of UrlInformationProviderInterface
    ),
);
````

I think we should just use `array`

File: DependencyInjection/Compiler/RegisterUrlInformationProviderPass.php [hunk 56]
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public function process(ContainerBuilder $container)
+    {
+        // feature not activated means nothing to add
+        if (!$container->hasDefinition('cmf_seo.sitemap.chain_provider')) {
+            return;
+        }
+
+        $chainProviderDefinition = $container->getDefinition('cmf_seo.sitemap.chain_provider');
+        $taggedServices = $container->findTaggedServiceIds('cmf_seo.sitemap.url_information_provider');
```
WouterJ: let's rename the tag to `cmf_seo.url_information_provider`, because (a) it's less to type so DX++ and (b) it's more generic, you never know if you encounter more use cases for this

Doctrine/Phpcr/SitemapUrlInformationProvider.php
ElectricMaxxx: was my storms fault, her does that refactoring for extract -> method :-)

Model/UrlInformation.php
ElectricMaxxx: :+1:

ElectricMaxxx: had been a simple mapping on the google spec for sitemaps, but i will change it

SeoPresentation.php
dbu: i was assuming that this is invalid, but actually its not:
http://3v4l.org/ZqIEm#v500 (maybe some stale php4 memory)

ok, lets document this BC break when extending the class.

Model/UrlInformation.php
ElectricMaxxx: and now i need to write a test for that model ... :-)

Model/UrlInformation.php
ElectricMaxxx: what should i do now?

Sitemap/ChainProvider.php
ElectricMaxxx: My Storm likes it :-)

Sitemap/ChainProvider.php
ElectricMaxxx: This will hopefully be a list of 2-5 items, and you wanna cache it? or am i wrong?

CHANGELOG.md
ElectricMaxxx: yea i am slow

Controller/SitemapController.php
ElectricMaxxx: shouldn't we chage that in all files on master?

Sitemap/ChainProvider.php
ElectricMaxxx: it is the same in the example you have shown

Sitemap/ChainProvider.php
ElectricMaxxx: i should empty the array every time i add a new one, really?

ElectricMaxxx:
6 to 200 lets go ...

SeoPresentation.php
ElectricMaxxx: happy:
https://github.com/symfony-cmf/SeoBundle/pull/196/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR5
?

Controller/SitemapController.php
WouterJ: yeah and all bundles...

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Changelog
 =========
 
+* **2015-14-02**: implement sitemap generation
+* **2015-14-02**: [BC BREAK] Relax method visibility on "\Symfony\Cmf\Bundle\SeoBundle\SeoPresentation"
```
WouterJ: I would say: "[BC BREAK] Changed method visibility from private to public of SeoPresentation#getSeoMetadata()"

DependencyInjection/CmfSeoExtension.php
```diff
@@ -196,6 +202,16 @@ private function loadAlternateLocaleProvider($config, ContainerBuilder $containe
                     array($container->getDefinition($alternateLocaleProvider))
                 )
             ;
+            try {
+                $container
+                    ->findDefinition('cmf_seo.sitemap.phpcr_provider')
+                    ->addMethodCall(
+                        'setAlternateLocaleProvider',
+                        array($container->getDefinition($alternateLocaleProvider))
```
WouterJ: we should use findDefinition here too (and on line 202 as well)

Sitemap/ChainProvider.php
```diff
@@ -0,0 +1,67 @@
...
+        ksort($this->sortedProviders);
+
+        foreach ($this->sortedProviders as $providers) {
+            $sortedProviders = array_merge($sortedProviders, $providers);
+        };
+        return$sortedProviders;
```
WouterJ: there should be an empty line before return statements and there should be a space between the return statement and the expression.

WouterJ:
As you changed the Configuration class, you should update the XML schema definition too. (201, let's stop here)

ElectricMaxxx:
@WouterJ you can live with #213 and #214 I will do them after this one.

Sitemap/ChainProvider.php
WouterJ: can't we make priority a feature of the tags and just do this sorting in the compiler?

DependencyInjection/CmfSeoExtension.php
ElectricMaxxx: this can't be actuall.

WouterJ: why not?

Sitemap/ChainProvider.php
ElectricMaxxx: @WouterJ you mean an extra attribute? oh yea that would make sense, otherwise the priority wouldn't ever be injected. :-)

ElectricMaxxx: done

Sitemap/ChainProvider.php
dbu: $this->sortedProviders would be the cache for the providers, needs to be reset each time a provider is added when doing caching

dbu: but see below

dbu:
imho this is ready to merge. @WouterJ is the xml schema correct now?

i would not do the caching of sorted providers - there will be 1, maybe two in normal cases, and they should only be used once. the overhead of fetching all urls of the system twice must be way more than re-sorting the providers so i think we should not bother.

@ElectricMaxxx can you squash one more time please?

ElectricMaxxx:
I will do the xml schema stuff in #213 with the missed schemas for alternate locale and error sites. i will squash it now.

WouterJ:
:white_check_mark:

dbu:
Party

Peekmo:
Hello,

Have you got an idea of when it will be available ?

Thank you

lsmith77:
in general it will be released with CMF 1.3 which is likely going to happen in march/april. then again, we could already release a new SeoBundle minor release before that .. at least from a quick skim, none of the changes in master require new versions of any dependency:
https://github.com/symfony-cmf/SeoBundle/compare/1.1.0...master#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780

dbu:
we should first get the xsd schema update merged. but apart that why not, lets create a release then.

dbu:
@Peekmo meanwhile, would you mind trying out in your project if the sitemap works as expected for you? `symfony-cmf/seo-bundle: 1.2.*@dev` should do the trick - if something else complains, add ` as 1.1.0` to the line to make composer think its version 1.1... not something to go online with, but to check if the feature works correctly.

WouterJ:
This means that the tab PR is not going to make it into SeoBundle 1.2 and that we should postpone tab support to CMF 1.4?

lsmith77:
no. not necessarily. we can do another minor release of the SeoBundle once we are ready to release CMF 1.3

dbu:
agreed, we can do a seo bundle 1.3 when we do the next general cmf release round and bump to sonata 2.4. minor versions don't cost us much.

PR #209, opened 2015-02-06
Allow disabling of content listener.

benglass:
We are working on integrating this into our CMS but would like to use our own content listener.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

## TO DO 

- [X] Add enable_content_listener boolean node
- [X] Update XML schema
- [X] Add tests
- [ ] Update docs

### DIFF
--- a/DependencyInjection/CmfSeoExtension.php
+++ b/DependencyInjection/CmfSeoExtension.php
@@ -79,8 +79,13 @@ class CmfSeoExtension extends Extension
             $this->loadSonataAdmin($config['sonata_admin_extension'], $loader, $container, $sonataBundles);
         }
 
-        if ($this->isConfigEnabled($container, $config['alternate_locale'])) {
-            $this->loadAlternateLocaleProvider($config['alternate_locale'], $container);
+        if ($config['enable_content_listener']) {
+
+            $loader->load('content-listener.xml');
+
+            if ($this->isConfigEnabled($container, $config['alternate_locale'])) {
+                $this->loadAlternateLocaleProvider($config['alternate_locale'], $container);
+            }
         }
 
         $errorConfig = isset($config['error']) ? $config['error'] : array();
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -92,6 +92,7 @@ class Configuration implements ConfigurationInterface
                         ->scalarNode('default_change_frequency')->defaultValue('always')->end()
                     ->end()
                 ->end()
+                ->booleanNode('enable_content_listener')->defaultTrue()->end()
             ->end()
         ;
 
--- a/Resources/config/schema/seo-1.0.xsd
+++ b/Resources/config/schema/seo-1.0.xsd
@@ -21,6 +21,7 @@
         <xsd:attribute name="content-key" type="xsd:string" />
         <xsd:attribute name="metadata-listener" type="enable_auto" />
         <xsd:attribute name="sonata-admin-extension" type="enable_auto" />
+        <xsd:attribute name="enable-content-listener" type="xsd:boolean" default="true" />
     </xsd:complexType>
 
     <xsd:complexType name="persistence">
--- a/Tests/Unit/DependencyInjection/CmfSeoExtensionTest.php
+++ b/Tests/Unit/DependencyInjection/CmfSeoExtensionTest.php
@@ -244,4 +244,25 @@ class CmfSeoExtensionTest extends AbstractExtensionTestCase
             'cmf_seo.sitemap.url_information_provider'
         );
     }
+
+    public function testDisableSeoContentListener()
+    {
+        $this->container->setParameter(
+            'kernel.bundles',
+            array(
+                'DoctrinePHPCRBundle' => true,
+                'CmfRoutingBundle' => true,
+            )
+        );
+        $this->load(array(
+            'persistence'   => array(
+                'phpcr' => true,
+            ),
+            'enable_content_listener' => false
+        ));
+
+        $this->assertContainerBuilderNotHasService(
+            'cmf_seo.event_listener.seo_content'
+        );
+    }
 }
--- a/Resources/config/content-listener.xml
+++ b/Resources/config/content-listener.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" ?>
+
+<container
+        xmlns="http://symfony.com/schema/dic/services"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <parameters>
+        <parameter key="cmf_seo.event_listener.seo_content.class">Symfony\Cmf\Bundle\SeoBundle\EventListener\ContentListener</parameter>
+    </parameters>
+
+    <services>
+        <service id="cmf_seo.event_listener.seo_content" class="%cmf_seo.event_listener.seo_content.class%">
+            <argument type="service" id="cmf_seo.presentation"/>
+            <argument>%cmf_seo.content_key%</argument>
+            <tag name="kernel.event_listener" event="kernel.request" method="onKernelRequest"/>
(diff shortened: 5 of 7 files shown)


dbu:
looks sane to me, and makes sense.

the only thing missing imo is a PR to adjust the doc, with a `.. since` to the configuration reference of the seo bundle.

lsmith77:
cc @ElectricMaxxx

DependencyInjection/Configuration.php
```diff
@@ -85,6 +85,9 @@ public function getConfigTreeBuilder()
                         ->scalarNode('enable_sibling_provider')->defaultValue(false)->end()
                     ->end()
                 ->end()
+                ->arrayNode('content_listener')
+                    ->canBeDisabled()
```
WouterJ: Instead, I think we should use a `booleanNode` and call it `use_content_listener`

WouterJ:
The XML schema needs to be updated too

DependencyInjection/Configuration.php
dbu: also possible. but if we ever want to have other configuration on content_listener, it will be unelegant. then again that is not too likely, so +1 for use_content_listener (without a content_listener array node)

benglass: This is actually how I originally wrote it (booleanNode) however my thought was we may want to allow specifying a service_id to use instead of the out of the box content listener so you could provide your own content listener and not have to rewrite the alternate locale provider portion. I'm perfectly happy to switch it to a booleanNode, just figured I'd provide my rational first

WouterJ: While we agree, I just wanted to point out that you can easily keep BC by keeping content_listener.enabled and use_content_listener in sync in case we want to add more config in the future.

WouterJ: Hmm, what about a content_listener setting (scalarNode) which could either be the service id or false/null to disable it?

benglass: The default value would be true which would cause it to use the existing one?

WouterJ: Default value would be thre service id of thre drfault listener.

benglass: I like this its more explicit

dbu: does the bundle need to know the service id of the default listener? if yes that would be the right thing to do. although, would that mean that if its the default value, SeoBundle loads the service and otherwise its assumes the application provides the service somewhere else? and, does the SeoBundle need to know the name of that listener, does it inject it elsewhere?

benglass: @dbu I think the way this is handled by FOSUserBundle where you can override the service id for the user manager is that the default service's id is like cmf_seo.content_listener_default and the dependency injection extension sets up cmf_seo.content_listener as an alias based on the configuration

dbu: alright, sounds like a plan!

ElectricMaxxx:
I think there a still some features missing in the xml schema :-(

ElectricMaxxx:
@benglass looks good to me,but can you add a test for both situations. Should be something like 
https://github.com/symfony-cmf/SeoBundle/blob/master/Tests/Unit/DependencyInjection/CmfSeoExtensionTest.php#L119-L122

dbu:
@benglass any chance to wrap this up? we think about tagging a new minor version of SeoBundle this week to get the sitemap generation feature in a stable release.

Tests/Unit/DependencyInjection/CmfSeoExtensionTest.php
```diff
@@ -121,11 +121,13 @@ public function testAlternateLocaleWithPhpcr()
             'Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr\AlternateLocaleProvider'
         );
 
-        $this->assertContainerBuilderHasServiceDefinitionWithMethodCall(
-            'cmf_seo.event_listener.seo_content',
-            'setAlternateLocaleProvider',
-            array($this->container->getDefinition('cmf_seo.alternate_locale.provider_phpcr'))
-        );
+        if ($this->container->has('cmf_seo.event_listener.seo_content')) {
```
ElectricMaxxx: Wyh are you doing this?

benglass: Because this PR adds the ability to remove the seo content listener, therefore this test needs to only assert that the event listener exists with this method call if the event listener exists

ElectricMaxxx: normally the alternate locale provider is injected into the SeoPresentation, when activated and with phpcr as implementation, this is done inside of the `CmfSeoExtension`. I would propose to add your changes there.

ElectricMaxxx: Do then you should use the config section above of the test and check for the existence by on or  many asserts. Cause i think your/my event listener will exist only, when there is a line in the configuration, right?

DependencyInjection/CmfSeoExtension.php
```diff
@@ -81,8 +81,17 @@ public function load(array $configs, ContainerBuilder $container)
             $this->loadSonataAdmin($config['sonata_admin_extension'], $loader, $container, $sonataBundles);
         }
 
-        if ($this->isConfigEnabled($container, $config['alternate_locale'])) {
-            $this->loadAlternateLocaleProvider($config['alternate_locale'], $container);
+        if ($config['content_listener']) {
```
benglass: @dbu @WouterJ @ElectricMaxxx 

This is the current state of the logic.

The default value of content_listener is cmf_seo.event_listener.seo_content_default
If this is the value of the config key then we load the content listener service definition (which is a tagged service, we dont want to do load file unless this is the listener in use or we would have multiple kernel request listeners)
If the user provides another service id, they are responsible for tagging it at as a kernel request listener. We could do this automatically but I believe it would have to be in compiler pass because the service may be from another bundle. I think we may be best off just allowing disabling of the content listener and provide doc instructions that you can just provide your own. The other thing about this PR is that the loadAlternateLocaleProvider functionality is only applied to the default seo content listener so the user would have to handle this in their service if they disable this listener.

At this point I am leaning towards just changing 'content_listener' to 'enable_content_listener' as a boolean and providing instructions for users to provide their own listeners. It reduces code complexity here.

WouterJ: Yeah, I thought we did things with the service that was configured as content listener. I agree with the tagging.

+1 for a booleanNode now

benglass:
@dbu @WouterJ @ElectricMaxxx 

I have now added tests and updated the xml schema. I do not have experience working with xsd files so can someone verify I did it correctly? The current schema file does not seem up to date with the other config values but I think you may have already fixed this in the master branch?

Doc PR will be coming but it would be great to get this merged in for a numbered release

Resources/config/schema/seo-1.0.xsd
```diff
@@ -39,4 +39,8 @@
             <xsd:enumeration value="canonical" />
         </xsd:restriction>
     </xsd:simpleType>
+
+    <xsd:simpleType name="enable_content_listener">
```
WouterJ: these changes should be removed and instead, you should add this between line 20 and 21 in this file:

````xsd
<xsd:attribute name="enable-content-listener" type="xsd:boolean" default="true" />
````

WouterJ:
Thanks @benglass! Apart from my XSD comment, it looks great.

Could you please also rebase this PR? Then Travis can run a build and we can merge if Travis is happy :)

benglass:
@WouterJ Squashed, rebase and awaiting tests

benglass:
@WouterJ @ElectricMaxxx @dantleech Tests passing ready for merge

WouterJ:
:+1:

dbu:
thanks a lot! created an issue so we don't forget the documentation: https://github.com/symfony-cmf/symfony-cmf-docs/issues/644

PR #218, opened 2015-02-17
(Sitemap) No duplicate locations from dynamic routing

Peekmo:
Hello,

With the dynamic router, the given resource has at least 2 routes (one for /cms/news/ID (for example) and one for /cms/routes/news/SLUG (for example too))

To avoid duplicate locations, I added the location as an array key. Don't know if that's a good thing (not very familiar with phpcr), this post will be more like an issue if my fix is dirty.

Regards,
Peekmo

### DIFF
--- a/Doctrine/Phpcr/SitemapUrlInformationProvider.php
+++ b/Doctrine/Phpcr/SitemapUrlInformationProvider.php
@@ -109,13 +109,14 @@ class SitemapUrlInformationProvider implements UrlInformationProviderInterface
             }
 
             try {
-                $urlInformationList[] = $this->computeUrlInformationFromSitemapDocument($document);
+                $urlInformation = $this->computeUrlInformationFromSitemapDocument($document);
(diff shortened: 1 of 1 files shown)


ElectricMaxxx:
@Peekmo is there a chance to cover the use case by a test?

Peekmo:
My bad..

I init my repository again and I don't have any duplicate value... Weird and I don't understand why, but this PR doesn't have any sense anymore.. Sorry..

Regards,
Peekmo

dbu:
max, do we only output the canonical urls or all urls in the sitemap?

ElectricMaxxx:
God question. Should do both, when having the right additional information on the tag.

ElectricMaxxx:
but i wil have a look in the evening. I think i would have to read a little bit about marking a route as canonical on sitemap, especially for google. Otherwise we should have flag in the sitemap config to enable/disable that behavior.

Issue #216, opened 2015-02-17
(Sitemap) Dependency failure if sitemap not enabled


Peekmo:
Hello,

If the sitemap is not enabled, there's an error because the service cmf_seo.sitemap.phpcr_provider depends on cmf_seo.sitemap.default_change_frequency which is not loaded if the sitemap is not enabled in configuration.

```
The service "cmf_seo.sitemap.phpcr_provider" has a dependency on a non-existent parameter "cmf_seo.sitemap.default_chan_frequency".
```

Regards,
Peekmo

dbu:
ups, thanks for spotting. @ElectricMaxxx so we probably should move that service into a specific configuration file after all.

ElectricMaxxx:
Oh. I will try to get aware of it and fetch that into a test too.

ElectricMaxxx:
@Peekmo is that bug still there?  Should have been fixed in #219 and in the final PR #225 Please reopen, when it still exists.

PR #219, opened 2015-02-17
fix it - call it default_change_frequency every where

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #216 
| License       | MIT

Yea found it, very strange that the configuration test and the webtest didn't get that problem. Changed the parameter name in the service injection only and kept it in configuration definition and its test. Ist the Definition created in the configuration test?


WouterJ:
> Ist the Definition created in the configuration test?

ConfigurationTest is only about the processing of the Configuration class, not about the handling of the extension.

ElectricMaxxx:
I think i ment the ExtensionTest, sorry. There where i try different configs and check got the existence of services and parameters.

dbu:
+1 - but i think the xml schema must be adjusted as well.

ElectricMaxxx:
done.

PR #224, opened 2015-02-18
French translation

Peekmo:
Hello,

I added the french translation for different labels.

Regards,
Peekmo


ElectricMaxxx:
@dbu task for you to check them? I don't understand them, or do we check those translations in generall?

WouterJ:
99% of the translations are based on trust I think.

ElectricMaxxx:
so then lets merge and thank @Peekmo  for his work.

Peekmo:
No problem, I'm french and I needed this translations ;) They are probably not the best translations ever, but they are better than nothing :)

PR #225, opened 2015-02-18
create named sitemaps

### LINKED ISSUE #212: Have a way to determine sitemap change frequency individually
The PHPCR SitemapUrlInformationProvider always assigns the default changefreq on the sitemap. Instead we could have an extractor for that information (e.g. based on class or an interface that lets tune this by document, or whatever custom logic).

See https://github.com/symfony-cmf/SeoBundle/pull/196/files#r24715055

### LINKED ISSUE #221: Move publish workflow checker to provider decorator
We should have a VotingProvider which decorates a `UrlInformationProviderInterface` implementation. This is then used to vote if the UrlInformation is shown. The publish workflow checker would then be one voter.

People might want to add more features, like ACL checks (as used by the kunstmaan guys https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/46 ).

This also means the `UrlInformation` class should have a new property: `$object` which contains the document/entity.

ElectricMaxxx:
fix #221, fix #212.

Introduce named sitemaps with its own defintions of default_change_frequency (later extracted by an guesser - just as fallback) and the templates.

Following tasks will be:
* [x] named sitemaps with configuration and schema
* [x] create a voters wich reduce the list of documents, based on the result of the provider
* [x] let guesser do the work of creating url information baed on the given documents (after vote)
* [x] each of them will be taged services with an attribute `sitemap` to define different for each sitemap
* [x] fix travis failing test
* [x] autocreate defautl sitemape configuration

Doc issue: https://github.com/symfony-cmf/symfony-cmf-docs/issues/681


Tests/Resources/Fixtures/config/config2.xml
```diff
@@ -11,5 +11,12 @@
 
     <error enable-parent-provider="true" enable-sibling-provider="true" />
 
-    <sitemap default-chan-frequency="never" enabled="true" />
+    <sitemap enabled="true">
+        <configuration default-chan-frequency="never" enabled="true" name="default">
+            <templates>
+                <template format="xml" template="test.xml" />
```
ElectricMaxxx: Not sure if the nested way is the right one. It was the easiest way of defining a schema and it is the way i expect the values in the `CmfSeoExtension`

ElectricMaxxx:
Or should i do all tasks in this PR? then i would change the list above into checkboxes.

DependencyInjection/Configuration.php
```diff
@@ -89,7 +89,28 @@ public function getConfigTreeBuilder()
                     ->addDefaultsIfNotSet()
                     ->canBeEnabled()
                     ->children()
-                        ->scalarNode('default_chan_frequency')->defaultValue('always')->end()
+
+                    ->end()
+                ->end()
+                ->arrayNode('sitemap')
```
dbu: i think you duplicated the sitemap arrayNode here

Controller/SitemapController.php
```diff
@@ -78,8 +86,8 @@ public function indexAction($_format)
         }
 
         $urls = $this->urlProvider->getUrlInformation();
```
dbu: i would put the urlProvider into the configuration as well. then i can have one provider for each sitemap.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -233,9 +233,18 @@ private function loadSitemapHandling($config, XmlFileLoader $loader, ContainerBu
...
-            $config['default_chan_frequency']
-        );
+        $container->setParameter($this->getAlias().'.sitemap.configurations', $config['configurations']);
+
+        foreach ($config['configurations'] as $key => $configuration) {
+            $container->setParameter($this->getAlias().'.sitemap.'.$key.'_configuration', $configuration);
+        }
+
+        // set a base default change frequency by now todo change that when votes respect sitemap by name
+        if (isset($config['configurations']['default']['default_change_frequency'])) {
```
dbu: should we autocreate the default sitemap if its not explicitly defined? 

and how do we handle default templates? we really need #217 now but i would say: if no templates are defined, we add the default ones for html and xml, otherwise we don't do anything extra and assume the user did what he wanted.

dbu:
i would prefer to do this in several PR so that we merge as soon as one part is good.

ElectricMaxxx:
You have had your chance, its to Late now. Got almost everything except the voter working in an half an Hour. :-)

dbu:
do you ever sleep? great job!

Tests/Resources/app/config/cmf_seo.yml
```diff
@@ -14,7 +14,13 @@ cmf_seo:
...
         enable_parent_provider: true
         enable_sibling_provider: true
-    sitemap: true
+    sitemap:
+        configurations:
+            default:
+                default_change_frequency: never
+                templates:
+                    xml: 'CmfSeoBundle:Sitemap:index.xml.twig'
+                    html: 'CmfSeoBundle:Sitemap:index.html.twig'
```
dbu: imho we should not need this explicitly when its the default values. we should just say sitemap: true and the default sitemap with these defaults is automatically defined. the Configuration could auto-add things. 

i can do a PR tomorrow to do this if you want.

DependencyInjection/CmfSeoExtension.php
ElectricMaxxx: this is a task i didn't get yesterday, found no real example tried to look into doctrine-bundle.

dbu: i think the best approach is to use the Configuration beforeValidation callbacks for this, not sure if it would even be possible in pure declarative code. i can have a go at that tomorrow if you want

ElectricMaxxx:
So ... This will fail for sure, as i still miss the voters. I need to do some refactoring in code duplications and re-write the tests for be base provider and guesser. But it should show the way it can work.

Controller/SitemapController.php
```diff
@@ -77,12 +94,19 @@ public function indexAction($_format)
             return new Response($text, 406);
         }
 
-        $urls = $this->urlProvider->getUrlInformation();
-        if (isset($this->templates[$_format])) {
-            return new Response($this->templating->render($this->templates[$_format], array('urls' => $urls)));
+        $urlInformations = array();
+        $documents = $this->routesForSitemapProvider->getDocumentsForSitemap($sitemap);
+        foreach ($documents as $document) {
+            $urlInformation = new UrlInformation();
```
ElectricMaxxx: I would implement the blocker of the voter right here.

Tests/Resources/app/config/cmf_seo.yml
ElectricMaxxx: :+1:

ElectricMaxxx:
@dbu i hope you understand the mechanism i tried to implement. Both, the guesser and the provider, are tagged services with an attribute for the priority and the sitemap name. Priority defaults to 0, the name to `default` at all. The providers are responsible to provide a list of documents/objects for the sitemap. The guesser will get each of them and change a given url information object. As botth, guesser and providers a priotisised you can chain them together in that way you want and can be bound to a specific sitemap. Thats the reason for both chains which implementing the same interface as the guesser or providers.

dbu:
sorry, i won't get much done today. but i looked over this and want to propose we do things a bit differently:

* the controller now became huge and is bound to the documentprovider - guesser paradigm
* i suggest that we keep the UrlInformationProvider as it was, and have an implementation with the code you are now adding to the controller method.
* i would call the thing you call document provider a "document loader".

ElectricMaxxx:
OK, pushing it to an extra service sound sane, so :+1:

Just the name for the loader should be something else. Cause what about 
simple objects somebody else is getting from the ORM?

So what about an ContentLoader?

Doctrine/Phpcr/SimpleUrlInformationGuesser.php
```diff
@@ -0,0 +1,89 @@
...
+
+        if ($this->alternateLocaleProvider) {
+            $collection = $this->alternateLocaleProvider->createForContent($object);
+            $urlInformation->setAlternateLocales($collection->toArray());
+        }
+
+        $seoMetadata = $this->seoPresentation->getSeoMetadata($object);
+        if (null !== $seoMetadata->getTitle()) {
+            $urlInformation->setLabel($seoMetadata->getTitle());
+            return $urlInformation;
```
WouterJ: shouldn't it always be returned?

WouterJ: hmm, no, it should never return anything :)

ElectricMaxxx: Exactly. Doing every change on the reference passed as an parameter.

Doctrine/Phpcr/SimpleUrlInformationGuesser.php
```diff
@@ -0,0 +1,89 @@
...
+use Symfony\Cmf\Bundle\SeoBundle\SeoPresentation;
+use Symfony\Cmf\Bundle\SeoBundle\Sitemap\UrlInformationGuesserInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * A base guesser to extract url information from a document persisted with the phpcr-odm.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+class SimpleUrlInformationGuesser implements UrlInformationGuesserInterface
```
WouterJ: I would split this guesser in 2 guessers (which you can then also give more descriptive names):

* `DefaultValueGuesser` (which should come at last, filling all open values with default values)
* `ObjectGuesser` (which executes the extractors and routing generators, etc.)

I also think this does not belong in the `Doctrine\Phpcr` namespace, as nothing in this class requires PHPCR-ODM stuff.

Doctrine/Phpcr/SimpleUrlInformationGuesser.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        $this->alternateLocaleProvider = $alternateLocaleProvider;
+    }
+
+    /**
+     * {@inheritDocs}
+     */
+    public function guessValues(UrlInformation $urlInformation, $object, $sitemap = 'default')
+    {
+        $urlInformation->setLocation($this->router->generate($object, array(), true));
```
WouterJ: we should add some checks first, to see if the value isn't already guessed by another guesser and if this guesser should overrule that value.

Actually, I think we should use a real guess implementation with a confidence. E.g:
````php
$values = array();

$values[] = new LocationGuess(
    $this->router->generate($object, array(), true),
    Guess::VERY_HIGH_CONFIDENCE
);

$values[] = new ChangeFrequencyGuess(
    $this->defaultChangeFrequency,
    Guess::LOW_CONFIDENCE
);

// ...
````

The value with the highest confidence is then set in the UrlInformation instance. Just like https://github.com/symfony/Form/blob/master/Guess/Guess.php

Comment (ElectricMaxxx) [inline Doctrine/Phpcr/SimpleUrlInformationGuesser.php]: you are completely right, was just my way of extracting the the lines into their destinations.

**Comment by ElectricMaxxx** [comment]
Yea the base structure is prepared now. Just need to re-implement the publich workflow checker as a voter now to get the former tests become green.

Would need some help with naming, but will clear them up a little bit first. 
Any hints for possible voters and guessers i should implement as default?

File: Controller/SitemapController.php [hunk 10]
```diff
@@ -64,9 +67,15 @@ public function __construct(
      *
      * @return Response
      */
-    public function indexAction($_format)
+    public function indexAction($_format, $sitemap)
```
dbu: should we do `$sitempa =  'default'`?

Controller/SitemapController.php
```diff
@@ -64,9 +67,15 @@ public function __construct(
      *
```
dbu: missing phpdoc for the sitemap parameter

DependencyInjection/Compiler/RegisterUrlInformationProviderPass.php
```diff
@@ -32,13 +32,34 @@ class RegisterUrlInformationProviderPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
+        $this->processTagsForService(
+            'cmf_seo.sitemap.document_provider',
```
dbu: why "document" - i think this should be called "object" here, its not only about phpcr-odm.

DependencyInjection/Compiler/RegisterUrlInformationProviderPass.php
```diff
@@ -32,13 +32,34 @@ class RegisterUrlInformationProviderPass implements CompilerPassInterface
...
+            'cmf_seo.sitemap.url_information_provider',
+            $container
+        );
+
+        $this->processTagsForService(
+            'cmf_seo.sitemap.guesser_provider',
```
dbu: should this not just be "guesser" instead of "guesser_provider"?

dbu:
wow, this got pretty big again :-)

i suggest we simplify to the following:

* Sitemap\Provider is a class with no interface, used by the controller. There is one provider per named sitemap. Its only api method is getSitemap() without argument.
* Sitemap\LoaderInterface has method load() that is supposed to return UrlInformation objects or things that can be converted to UrlInformation objects. The SitemapProvider accepts more than one loader and implements the chaining.
* Sitemap\VoterInterface has a method vote() to decide whether a thing returned by the loader should go into the sitemap.
* Sitemap\GuesserInterface has a method guess to enrich a UrlInformation object from the data. there can be several guessers. A guesser SHOULD NOT update a field that is already set. the priorities are thus handled by registering the guessers. I don't see how we could go into more detail here (confidence) without become even more complicated which would seem over the top for this usecase imho. if you really think its worth it, the approach wouter proposed for adding guesses seems right to me.

loaders and guessers are tagged sitemap.loader and sitemap.guesser . you can specify an extra attribute on the tag called sitemap=xxx to limit them to specific sitemaps. that attribute value can be comma separated. they also have a priority attribute.

as for implementations, there is a phpcr specific loader, the published voter tying in the CoreBundle publish workflow and guessers using the seo extractor and a default value guesser.

i want to propose not to pass sitemap names around at all, so you control things by setting up different services on the different SitemapProvider instances. but i am not 100% sure if that is not becoming inconvenient for loaders or voters. having only one provider would complicate things further however, and having both leaves too many different ways to do the same, imho.

wdyt?

ElectricMaxxx:
The names sounds sane to me. What i do not understand is the comma separated attribute. You whant something like `sitemap="1,my-sitemap"` instead of `priority="1" sitemap"my-sitemap"`?

I would use the different sitemap approach in all cases, so it will be easy for one (no attributes) and yea for more than one sitemap it gonna be hard, but a page with more than one sitemap will be complicated at all.

I think one provider per sitemap gonna be hard to, cause who wants to inject the right one into the controller? I think have the provider stuff out of the controller is cool, but i don't really wanna have some hard configuration-hacking to have the matching provider inside of the controller.

I will clean up the names now. So we will get a better view later.

WouterJ:
@ElectricMaxxx David menas `sitemap="sitemap1, sitemap2, ..." priority="99"`

dbu:
yep, exactly thats what i meant with comma separating.

you inject a configuration per sitemap name anyways, for the templates. the provider should just be one field in that configuration. i think this is the most elegant aproach to tweak when to use which voters or providers (or guessers, if they are different). it means that we need to copy the service definition for each sitemap name, as per the configuration of the templates. we should also validate the tag to only reference existing sitemaps

Doctrine/Phpcr/SimpleUrlInformationGuesser.php
ElectricMaxxx: That is a slightly different solution to the priority thing on the attributes, right?

ElectricMaxxx:
Cleared the names of classes, services and tags. So let the big refactoring began. Tests are green again. So the parts are working as before.

Doctrine/Phpcr/SimpleUrlInformationGuesser.php
dbu: yeah, this would be the all-out complex solution. it is more powerful and puts less logic into single guessers, as they would not need to check if a field is set already or not. but it adds yet more things. but we could go that way. not sure if there is any reusable guesser infrastructure we could take or whether we would need to completely implement things here.

ElectricMaxxx: @WouterJ got a little dependency problem with that. 

Normaly we try to implement a default behavior and some extra stuff for stuff that dependens on phpcr. For example when phpcr is chosen the `PublicWorkflowUrlInformationVoter` starts working and takes care that public content is shown on sitemap only.
But what should i do for your suggestion of guessers? There are some values that are easy: `label` (SeoMetdata), but the router needs phpcr to return a valid url based on the content object - null would be a nice url.
So what should i do? One guesser, which works when phpcr is chosen? but how would you customize that? I can't work with the chain filled by tagged services, as each one need its own value.

dbu: in the interest of getting something done that works, rather than build a huge thing:

could we simply build this directly as a guesser and not use any chain for the first iteration? as long as we follow the interfaces and allow to configure the service, adding a chain later on should be possible without BC break.

ElectricMaxxx: So you would suggest to have just ONE guesser, which Is able to extract the values and implements an interface, Right? So i would remove the chain and tagging stuff for the guesser. We wouldn't break anything by replacing the chain By @wouterJ 's Suggestion or something else later, Right? Sounds easy. I like your Plan.

dbu: yes, that is my suggestion.

building a full guesser infrastructure sounds huge for just this feature. actually makes me wonder if the symfony form type guesser infrastructure could be extracted into its own component to make it reusable. @webmozart what do you think about that? could this be something for symfony3, building a generic guesser component?

ElectricMaxxx: Yea @dbu i think this would be a nice idea. I would need that guesser component in an actual project in my Company too. 

I will try to finish that PR tonight.

ElectricMaxxx: But @dbu would have a dependecy confusion again:

```
        $urlInformation->setLocation($this->router->generate($object, array(), true));
        $urlInformation->setChangeFrequency($this->defaultChangeFrequency);

        if ($this->alternateLocaleProvider) {
            $collection = $this->alternateLocaleProvider->createForContent($object);
            $urlInformation->setAlternateLocales($collection->toArray());
        }

        $seoMetadata = $this->seoPresentation->getSeoMetadata($object);
        if (null !== $seoMetadata->getTitle()) {
            $urlInformation->setLabel($seoMetadata->getTitle());
            return $urlInformation;
        }
``` 

would be the current code for a simple global guesser for the url information. But there are at least 2 properties, which are created by a service which depends on `PHPCR`: location (router won't work) and alternate locale. Not to set the alternate locale for a non existing service would be ok, but the location? 
At the end the complete guesser depends on `PHCR`, but we won't be able to skip the guesser itself inside of the sitemap controller, as there would be no content at all. So the controller itself depends on `PHPCR`.

ElectricMaxxx:
So ...
What is missing now?

* multiple sitemap names: `sitemap="name1, name2,.."
*  beforeValidation callbacks for configuration

Controller/SitemapController.php
```diff
@@ -25,48 +26,57 @@
...
+        Provider $informationProvider,
         EngineInterface $templating,
-        array $templates
+        array $configurations
     ) {
-        $this->urlProvider = $provider;
         $this->templating = $templating;
-        $this->templates = $templates;
+        $this->configurations = $configurations;
+        $this->informationProvider = $informationProvider;
```
dbu: can we call this variable `sitemapProvider`?

DependencyInjection/CmfSeoExtension.php
```diff
@@ -187,26 +188,30 @@ private function loadPhpcr($config, XmlFileLoader $loader, ContainerBuilder $con
...
+                        'setAlternateLocaleProvider',
+                        array($alternateLocaleProviderDefinition)
+                    );
+                $container
+                    ->findDefinition('cmf_seo.sitemap.guesser')
+                    ->addMethodCall(
+                        'setAlternateLocaleProvider',
+                        array($alternateLocaleProviderDefinition)
+                    );
+            } catch(InvalidArgumentException $e) {
```
dbu: can we remove this try-catch again?

Doctrine/Phpcr/PublicWorkflowUrlInformationVoter.php
```diff
@@ -0,0 +1,44 @@
...
+     * A voter should decide whether a content object should be exposed on a sitemap.
+     *
+     * @param object $content
+     * @param string $sitemap
+     *
+     * @return bool
+     */
+    public function exposeOnSitemap($content, $sitemap = 'default')
+    {
+        return $this->publishWorkflowChecker->isGranted(array(PublishWorkflowChecker::VIEW_ATTRIBUTE), $content);
```
dbu: no need to wrap the attribute in an array, the checker does that for you

Doctrine/Phpcr/PublicWorkflowUrlInformationVoter.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->publishWorkflowChecker = $publishWorkflowChecker;
+    }
+    /**
+     * A voter should decide whether a content object should be exposed on a sitemap.
+     *
+     * @param object $content
+     * @param string $sitemap
+     *
+     * @return bool
+     */
```
dbu: use `{@inheritDoc}` here

Doctrine/Phpcr/PublicWorkflowUrlInformationVoter.php
```diff
@@ -0,0 +1,44 @@
...
+namespace Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr;
+
+use Symfony\Cmf\Bundle\CoreBundle\PublishWorkflow\PublishWorkflowChecker;
+use Symfony\Cmf\Bundle\SeoBundle\Sitemap\VoterInterface;
+use Symfony\Component\Security\Core\SecurityContextInterface;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class PublicWorkflowUrlInformationVoter implements VoterInterface
```
dbu: lets call this PublishWorkflowUrlInformationVoter and not Public..

Doctrine/Phpcr/PublicWorkflowUrlInformationVoter.php
```diff
@@ -0,0 +1,44 @@
...
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2015 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr;
```
dbu: this voter is not specific to phpcr or doctrine in general.

Sitemap/GuesserInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+interface GuesserInterface
+{
+    /**
+     * Guesses url information values from a given object.
```
dbu: lets also say that this method updates $urlInformation with new values.

dbu:
this needs a rebase.
travis is very unhappy about some BC break with sonata blocks - we probably need to restrict what version we are compatible with.

dbu:
i will be away for another week, so wont have time very soon to do something. if nobody gets to it before, i can then try to contribute the configuration thing for the default sitemap configuration. 
what do you mean by the "multiple sitemap names"?

DependencyInjection/CmfSeoExtension.php
ElectricMaxxx: no we can't. 
There are situations, where the service `cmf_seo.sitemap.guesser` isn't available - alternate locale actived, sitmap not. That would cause an exception for the `findDefintion()` method.

dbu: but then we should do $container->hasDefinition or whatever its called to check, if we can't even set a flag in the class to know whether the guesser should be available or not. silently discarding an exception is quite risky, and should not be needed for this case.

ElectricMaxxx: thats a thing @WouterJ suggested to do: use `findDefintion()` at all don't use `hasDefintion()`.

ElectricMaxxx:
Any ideas what goes wrong on travis?

DependencyInjection/CmfSeoExtension.php
WouterJ: I would propose to use `if ($container->has('...')) { $something = $container->findDefinition('...'); }`

DependencyInjection/CmfSeoExtension.php
```diff
@@ -208,23 +211,25 @@ private function loadAlternateLocaleProvider($config, ContainerBuilder $containe
...
-        }
-        if ($this->phpcrProviderEnabled) {
-            $container
-                ->findDefinition('cmf_seo.sitemap.phpcr_provider')
-                ->addMethodCall(
-                    'setAlternateLocaleProvider',
-                    array($alternateLocaleProviderDefinition)
-                )
-            ;
+        if ($alternateLocaleProvider) {
```
dbu: we already returned above if $alternateLocaleProvider is not set, so this check can be removed and indention be reverted.

Sitemap/AbstractChain.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Sitemap;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
```
dbu: can you please add a phpdoc here what this is?

Sitemap/Guesser.php
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Sitemap;
+
+use Symfony\Cmf\Bundle\SeoBundle\AlternateLocaleProviderInterface;
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+use Symfony\Cmf\Bundle\SeoBundle\SeoPresentation;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * A chain to register all url information guesser by its priority and preferred sitemap.
```
dbu: i don't see a chain in here. copy-paste?

Sitemap/GuesserInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Sitemap;
+
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
```
dbu: can you please add a short explanation what this interface does?

Sitemap/AbstractChain.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $this->items[$sitemap][$priority][] = $item;
+    }
+
+    /**
+     * @param $sitemap
+     *
+     * @return array
+     */
+    protected function getSortedItemsBySitemap($sitemap)
```
dbu: i think this should be called "getSortedItemsForSitemap". by sitemap made me think that the items are sorted by sitemap. but we only ever get items for one specific sitemap

Sitemap/AbstractChain.php
```diff
@@ -0,0 +1,56 @@
...
+            $this->items[$sitemap][$priority] = array();
+        }
+
+        $this->items[$sitemap][$priority][] = $item;
+    }
+
+    /**
+     * @param $sitemap
+     *
+     * @return array
```
dbu: can you explain a bit what this does?

Sitemap/AbstractChain.php
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
+ */
+abstract class AbstractChain
+{
+    /**
+     * The list of guessers with sitemap name and priorities as key.
+     *
+     * @var
```
dbu: this is an array of objects.

Sitemap/Voter.php
```diff
@@ -0,0 +1,27 @@
...
+ * A chain as a decorator for all tagged voter services.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class Voter extends AbstractChain implements VoterInterface
+{
+    /**
+     * {@inheritDocs}
+     *
+     * First comes first - The first negative voter decides.
```
dbu: in terms of symfony security, this would be called "unanimous": every voter can veto a content.

Sitemap/Voter.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Sitemap;
+
+/**
+ * A chain as a decorator for all tagged voter services.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class Voter extends AbstractChain implements VoterInterface
```
dbu: don't remember if we discussed this before, but could we remove the whole voting implementation and use our own instance of `AccessDecisionManager` with strategy `unanimous` and have the voters implement the symfony security `VoterInterface`?

i think the security component does a good job at modeling the whole voting process.

dbu: the publish workflow in core bundle is doing this successfully, could be an example to look at how to do this.

dbu:
i think what we saw on travis is related to changes in the testing component and phpcr-odm. i just restarted to see if that helps.

Sitemap/Voter.php
ElectricMaxxx: uhhh, ohhh this would make my abstraction for the chain useless, as there will be one chain afterwards.

dbu: which would mean less code to maintain ;-)

WouterJ:
The tests are failing because PHP code coverage needs to be a stable version in order to support HHVM. Add `"phpunit/php-code-coverage": "@stable"` to `require-dev` to solve the problem.

Sitemap/Voter.php
ElectricMaxxx: @dbu i can't really follow your ideas with the `VoterInterface`. The `attribute` can be like:
```
[ sitemap => 'sitemap1, sitemap2, ..']
```
The object is the document/content to vote for, but what should i use as a token? I have got no security context here, is there an empty token (new) that i should use?

dbu: lets drop my idea :-/

ElectricMaxxx: So i would just fix tests and then merge?

Am 16.07.2015 um 08:55 schrieb David Buchmann:

ElectricMaxxx:
@dbu i gave you some help:
- rebased the branch
- reverted the voter changes i did
- test should be green in general

but there is an method signature fail in `\Sonata\AdminBundle\Block\AdminListBlockService` localy. I fixed that localy to get the tests green, but we should be aware of that.

dbu:
thanks. i started some refactoring, will do a PR against this PR soonish.

ElectricMaxxx:
And please don't forget the task with the configuration: singel/multiple sitemaps. Thanks a Lot.

ElectricMaxxx:
I will hopefully have some Time to review the PR tonight.

DependencyInjection/Configuration.php
```diff
@@ -102,18 +102,31 @@ public function getConfigTreeBuilder()
                     ->end()
                 ->end()
                 ->arrayNode('sitemap')
```
ElectricMaxxx: @dbu didn't we want to change/extend this? You wanted to use single configurations too, right?

dbu:
okay, now we have more stuff. what was the task with the multiple sitemaps configuration again? is that covered or are there more things to consider?

was it only the failing tests or is there other things that remain to be done?

ElectricMaxxx:
I mean your own commit:

https://github.com/symfony-cmf/SeoBundle/pull/225#discussion_r24973222
:-)

The tests had been fixed (before your PR)

ElectricMaxxx:
yea the tests are still green, so the configuration thing is missing only.

DependencyInjection/Configuration.php
dbu: ah right, now i remember what we said. will give that a try tonight.

ElectricMaxxx:
rebased and squashed, merge when green?

dbu:
:+1: - please remove the [WIP] from the title :-)

dbu:
\o/

PR #226, opened 2015-02-21
adjust content_listener configuration

dbu:
i just realized that there is a content_key configuration that is also only relevant to the content_listener. i propose we consolidate that. for the 1.1 branch i just change what was committed now.

once this is good, i will update master and move the configuration for the key to the content_listener section, with BC code to still allow the old value.

doc: https://github.com/symfony-cmf/symfony-cmf-docs/pull/649

TODO: adjust changelog


WouterJ:
As you didn't release 1.1.1 yet, I would know vote for reverting this in 1.1 and doing things the correct way in master.

dbu:
:-( but i think you are right. created #227 - if we merge that, i will port this back to master.

dbu:
closing in favor of #228

PR #228, opened 2015-02-24
cleanup content_listener configuration

dbu:
realized that we should create a content_listener section.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | travis
| Fixed tickets | #226
| License       | MIT
| Doc PR        | https://github.com/symfony-cmf/symfony-cmf-docs/pull/649

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 Changelog
 =========
 
-* **2015-14-02**: implement sitemap generation
-* **2015-14-02**: [BC BREAK] Changed method visibility from private to public of SeoPresentation#getSeoMetadata()
+* **2015-02-24**: Configuration for content_key moved to content_listener section, and its now possible to disable
+  The content listener by setting cmf_seo.content_listener.enabled: false
+* **2015-02-14**: implement sitemap generation
+* **2015-02-14**: [BC BREAK] Changed method visibility from private to public of SeoPresentation#getSeoMetadata()
 * **2014-10-04**: Custom exception controller for error handling
 
 1.1.0-RC3
--- a/DependencyInjection/CmfSeoExtension.php
+++ b/DependencyInjection/CmfSeoExtension.php
@@ -31,6 +31,16 @@ class CmfSeoExtension extends Extension
     private $defaultAlternateLocaleProviderId;
 
     /**
+     * @var bool Whether the content listener is loaded
+     */
+    private $contentListenerEnabled = false;
+
+    /**
+     * @var bool Whether the phpcr sitemap provider is loaded
+     */
+    private $phpcrProviderEnabled = false;
+
+    /**
      * {@inheritDoc}
      */
     public function load(array $configs, ContainerBuilder $container)
@@ -43,16 +53,6 @@ class CmfSeoExtension extends Extension
         $loader->load('extractors.xml');
 
         $this->loadSeoParameters($config, $container);
-
-        if (empty($config['content_key'])) {
-            if (! class_exists('Symfony\Cmf\Bundle\RoutingBundle\Routing\DynamicRouter')) {
-                throw new \RuntimeException('You need to set the content_key when not using the CmfRoutingBundle DynamicRouter');
-            }
-            $contentKey = DynamicRouter::CONTENT_KEY;
-        } else {
-            $contentKey = $config['content_key'];
-        }
-        $container->setParameter($this->getAlias() . '.content_key', $contentKey);
 
         $sonataBundles = array();
         if ($this->isConfigEnabled($container, $config['persistence']['phpcr'])) {
@@ -79,20 +79,19 @@ class CmfSeoExtension extends Extension
             $this->loadSonataAdmin($config['sonata_admin_extension'], $loader, $container, $sonataBundles);
         }
 
-        if ($config['enable_content_listener']) {
-
-            $loader->load('content-listener.xml');
-
-            if ($this->isConfigEnabled($container, $config['alternate_locale'])) {
-                $this->loadAlternateLocaleProvider($config['alternate_locale'], $container);
-            }
-        }
-
         $errorConfig = isset($config['error']) ? $config['error'] : array();
         $this->loadErrorHandling($errorConfig, $container);
 
         if ($this->isConfigEnabled($container, $config['sitemap'])) {
-            $this->loadSitemapHandling($config['sitemap'], $loader, $container);
+            $this->loadSitemapHandling($config['sitemap'], $loader, $container, $config['persistence']);
+        }
+
+        if ($this->isConfigEnabled($container, $config['content_listener'])) {
+            $this->loadContentListener($config['content_listener'], $loader, $container);
+        }
+
+        if ($this->isConfigEnabled($container, $config['alternate_locale'])) {
+            $this->loadAlternateLocaleProvider($config['alternate_locale'], $container);
         }
     }
 
@@ -178,7 +177,15 @@ class CmfSeoExtension extends Extension
         }
 
         $loader->load('matcher_phpcr.xml');
-        $loader->load('phpcr-sitemap.xml');
+    }
+
+    private function loadContentListener(array $config, XmlFileLoader $loader, ContainerBuilder $container)
+    {
+        $container->setParameter($this->getAlias() . '.content_key', $config['content_key']);
+
+        $loader->load('content-listener.xml');
+
+        $this->contentListenerEnabled = true;
     }
 
     /**
@@ -197,21 +204,27 @@ class CmfSeoExtension extends Extension
             ? $this->defaultAlternateLocaleProviderId
             : $config['provider_id'];
 
+        if (!$alternateLocaleProvider) {
+            return;
+        }
 
-        if ($alternateLocaleProvider) {
-            $alternateLocaleProviderDefinition = $container->findDefinition($alternateLocaleProvider);
+        $alternateLocaleProviderDefinition = $container->findDefinition($alternateLocaleProvider);
+        if ($this->contentListenerEnabled) {
             $container
                 ->findDefinition('cmf_seo.event_listener.seo_content')
                 ->addMethodCall(
                     'setAlternateLocaleProvider',
                     array($alternateLocaleProviderDefinition)
                 );
+        }
+        if ($this->phpcrProviderEnabled) {
             $container
                 ->findDefinition('cmf_seo.sitemap.phpcr_provider')
                 ->addMethodCall(
                     'setAlternateLocaleProvider',
                     array($alternateLocaleProviderDefinition)
-                );
+                )
+            ;
         }
     }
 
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 Changelog
 =========
 
-* **2015-14-02**: implement sitemap generation
-* **2015-14-02**: [BC BREAK] Changed method visibility from private to public of SeoPresentation#getSeoMetadata()
+* **2015-02-24**: Configuration for content_key moved to content_listener section, and its now possible to disable
+  The content listener by setting cmf_seo.content_listener.enabled: false
+* **2015-02-14**: implement sitemap generation
+* **2015-02-14**: [BC BREAK] Changed method visibility from private to public of SeoPresentation#getSeoMetadata()
```
dbu: these dates are in YYYY-MM-DD format, not sure if YYYY-DD-MM is a thing even in the US :-)

DependencyInjection/CmfSeoExtension.php
```diff
@@ -242,5 +250,9 @@ private function loadSitemapHandling($config, XmlFileLoader $loader, ContainerBu
             $this->getAlias().'.sitemap.default_change_frequency',
             $config['default_change_frequency']
         );
+
+        if ($this->isConfigEnabled($container, $persistence['phpcr'])) {
+            $loader->load('phpcr-sitemap.xml');
+        }
```
dbu: moving this here, as otherwise we load this bit even when sitemaps are not enabled, leading to missing default_change_frequency parameter.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -205,13 +209,17 @@ private function loadAlternateLocaleProvider($config, ContainerBuilder $containe
...
-                );
-            $container
-                ->findDefinition('cmf_seo.sitemap.phpcr_provider')
-                ->addMethodCall(
-                    'setAlternateLocaleProvider',
-                    array($alternateLocaleProviderDefinition)
-                );
+                )
+            ;
+            if ($container->hasDefinition('cmf_seo.sitemap.phpcr_provider')) {
```
dbu: this may or may not be loaded.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -79,9 +70,8 @@ public function load(array $configs, ContainerBuilder $container)
             $this->loadSonataAdmin($config['sonata_admin_extension'], $loader, $container, $sonataBundles);
         }
 
-        if ($config['enable_content_listener']) {
-
-            $loader->load('content-listener.xml');
+        if ($this->isConfigEnabled($container, $config['content_listener'])) {
+            $this->loadContentListener($config['content_listener'], $loader, $container);
 
             if ($this->isConfigEnabled($container, $config['alternate_locale'])) {
```
dbu: @ElectricMaxxx does it make sense that this is only loaded when the content listener is enabled? could we not need alternate locale for other things too, like the sitemap?

Resources/config/schema/seo-1.0.xsd
```diff
@@ -18,10 +18,9 @@
         <xsd:attribute name="title" type="xsd:string" />
         <xsd:attribute name="description" type="xsd:string" />
         <xsd:attribute name="original-route-pattern" type="original_route_pattern" />
-        <xsd:attribute name="content-key" type="xsd:string" />
```
dbu: @WouterJ do we need to bump the schema version? or keep this in the deprecated location too, as we have BC to still handle it in the Configuration class?

CHANGELOG.md
benglass: @dbu yes this is fine we havent started using this just yet so no changes will be necessary on our end

DependencyInjection/CmfSeoExtension.php
```diff
@@ -79,20 +69,19 @@ public function load(array $configs, ContainerBuilder $container)
...
         if ($this->isConfigEnabled($container, $config['sitemap'])) {
-            $this->loadSitemapHandling($config['sitemap'], $loader, $container);
+            $this->loadSitemapHandling($config['sitemap'], $loader, $container, $config['persistence']);
+        }
+
+        if ($this->isConfigEnabled($container, $config['content_listener'])) {
+            $this->loadContentListener($config['content_listener'], $loader, $container);
+        }
+
+        if ($this->isConfigEnabled($container, $config['alternate_locale'])) {
```
dbu: moving this out of enable content listener, as the alternate locale provider is also used by sitemap.

dbu:
okay, green this time. @ElectricMaxxx @benglass can you check if you agree?

and @WouterJ wdyt about BC with the xml schema?

DependencyInjection/CmfSeoExtension.php
```diff
@@ -178,7 +167,21 @@ private function loadPhpcr($config, XmlFileLoader $loader, ContainerBuilder $con
         }
 
         $loader->load('matcher_phpcr.xml');
-        $loader->load('phpcr-sitemap.xml');
+    }
+
+    private function loadContentListener(array $config, XmlFileLoader $loader, ContainerBuilder $container)
+    {
+        if (empty($config['content_key'])) {
+            if (! class_exists('Symfony\Cmf\Bundle\RoutingBundle\Routing\DynamicRouter')) {
```
WouterJ: you should remove the space after `!`

Resources/config/schema/seo-1.0.xsd
WouterJ: if we still support the old way in the Configuration class, we should also support it in the schema.

WouterJ: btw, if there are BC breaks in the schema, I'm +1 for bumping the version of it

WouterJ: another solution is to have 2 versions: 1.0 being the old one, 1.1 the new syntax. When going to 2.0, 1.0 is then going to be removed.

dbu:
fixed the CS issue and made the schema BC. imho its too early to go for a 2.0 schema - we might have other changes when we do a 2.0 so we should keep that option open.

dbu:
@ElectricMaxxx ok to merge this one? then i could start working on the wrap-up for the named sitemaps.

benglass:
@dbu @ElectricMaxxx 

This looks good to me. Looking forward to seeing this a release so we can start using seo bundle!

:+1:

dbu:
@ElectricMaxxx @WouterJ any last feedback or can we merge this?

WouterJ:
:+1:

dbu:
@ElectricMaxxx good to merge?

DependencyInjection/CmfSeoExtension.php
```diff
@@ -197,21 +200,27 @@ private function loadAlternateLocaleProvider($config, ContainerBuilder $containe
...
             : $config['provider_id'];
 
+        if (!$alternateLocaleProvider) {
+            return;
+        }
 
-        if ($alternateLocaleProvider) {
-            $alternateLocaleProviderDefinition = $container->findDefinition($alternateLocaleProvider);
+        $alternateLocaleProviderDefinition = $container->findDefinition($alternateLocaleProvider);
+        if ($container->hasDefinition('cmf_seo.event_listener.seo_content')) {
```
ElectricMaxxx: You should use findDefintion, as i did it to make @WouterJ happy.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -197,21 +200,27 @@ private function loadAlternateLocaleProvider($config, ContainerBuilder $containe
...
             $container
                 ->findDefinition('cmf_seo.event_listener.seo_content')
                 ->addMethodCall(
                     'setAlternateLocaleProvider',
                     array($alternateLocaleProviderDefinition)
                 );
+        }
+        if ($container->hasDefinition('cmf_seo.sitemap.phpcr_provider')) {
```
ElectricMaxxx: I thought i have done it her too?!? Just a short try-catch around them all.

ElectricMaxxx:
Except the usage of has/findDefinition it looks fine to me.

DependencyInjection/CmfSeoExtension.php
dbu: i do that below. if i don't check first, i get an error. i could check the value on $config but that seems more complicated to me and has no gain as i still use the service name below.

WouterJ: You have to use `has` here instead of `hasDefinition`

DependencyInjection/CmfSeoExtension.php
dbu: why try-catch? that would abort me.

hm, there is one reason to check $config: we detect when the service is expected to be loaded but not found (typo)

ElectricMaxxx: cause `findDefintion()` throws an exception.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -197,21 +204,27 @@ private function loadAlternateLocaleProvider($config, ContainerBuilder $containe
...
+        $alternateLocaleProviderDefinition = $container->findDefinition($alternateLocaleProvider);
+        if ($this->contentListenerEnabled) {
             $container
                 ->findDefinition('cmf_seo.event_listener.seo_content')
                 ->addMethodCall(
                     'setAlternateLocaleProvider',
                     array($alternateLocaleProviderDefinition)
                 );
+        }
+        if ($this->phpcrProviderEnabled) {
```
dbu: decided to be explicit here. this is more robust to discover bugs when we mix up service names or the like.

dbu:
ok, updated. good to merge?

Resources/config/schema/seo-1.0.xsd
```diff
@@ -12,16 +12,17 @@
...
             <xsd:element name="error" type="error_handling" minOccurs="0" />
             <xsd:element name="sitemap" type="sitemap" minOccurs="0" />
+            <xsd:element name="content-listener" type="content_listener" minOccurs="0" />
         </xsd:sequence>
 
         <xsd:attribute name="translation-domain" type="xsd:string" />
         <xsd:attribute name="title" type="xsd:string" />
         <xsd:attribute name="description" type="xsd:string" />
         <xsd:attribute name="original-route-pattern" type="original_route_pattern" />
+        <!-- the content-key attribute should be configured in the content-listener element but this attribute is kept for BC !-->
```
WouterJ: `<!-- Deprecated as of version 1.1, to be removed in version 2.0. Use the content-listener element instead.`

dbu:
updated the deprecation comment, thanks for the suggestion wouter.

dbu:
@benglass so this is merged. i hope you can now start to use this :-)

benglass:
@dbu Thanks!

Issue #222, opened 2015-02-18
Extract Sitemap to an own bundle


ElectricMaxxx:
Based on the discussions on #221 and #212 and the PR's following the introducion of the sitemap behavior i would suggest to extract its code and behavior to an bundle called `CmfSitemapBundle`. The functionality is decoupled from the Seo functionality at all and will be extended by the issues mentioned above. 

So should we do it?

dantleech:
I think that makes sense .. +1

dbu:
hm. if it is possible to disable the rest of the SEO stuff, then what would it hurt to have them in the same bundle? if we separate them, we add overhead and need to duplicate things like the extractor mechanism. and no, having a CmfExtractor library seems not the way to go to me. i think we are in a dilemma between clear responsibility and too many bundles. then again, CmfSitemapBundle with no dependency on sonata sounds very clear what it is about.

what are the overlaps we have? extract description and title? maybe we can make that an optional dependency with the option to supply your own code. or could we make the SitemapBundle *depend* on the SeoBundle for these features?

ElectricMaxxx:
@dbu  we need to write own sitemap extractors/enhancers/guessers i think. We should couple both kind of extractors.

dbu:
just had this idea of looking into packagist: https://packagist.org/search/?q=sitemap-bundle

there is like 10 sitemap bundles. some even build on doctrine. maybe we can just leverage that? a pitty for all the work already invested, but we should seriously look at those and think if we can simply write a phpcr-odm integration for one of those.

WouterJ:
There aren't much sitemap bundles that are decoupled out there. Many have requirements on core bundles from their systems, some are way beyond the sitemap scope, most of them require only XML format.

I don't think we can use one of them (please note that we had done some research before the PR started on this topic).

I also don't believe we should make it into a seperate bundle, it's all part of the SEO topic. Yeah, they are decoupled, but if we make this bundle able to disable/enable each seperate feature, I think it's better to keep them in one repository.

ElectricMaxxx:
I will let the vote open until the evening, and start to implement #221 and #212  then either in a new bundle or here. 
votes atm:
:+1: 2x (me too)
:-1: 1x
use an other bundle (no option before): 1x

btw: is there no chance to da a good poll in github?

dbu:
@ElectricMaxxx tried http://poll.gitrun.com/ but looks rather pointless to me.

[...]

PR #236, opened 2015-07-14
issue#235 Made it set lastModification in xml sitemap

WillBBP:
I just added the few lines below to the method computeUrlInformationFromSitemapDocument() in Doctrine/Phpcr/SitemapUrlInformationProvider.php so that where the sitemap document has a method getLastModification(), this method is used to set lastModification in the $urlInformation.

### DIFF
--- a/Doctrine/Phpcr/SitemapUrlInformationProvider.php
+++ b/Doctrine/Phpcr/SitemapUrlInformationProvider.php
@@ -157,6 +157,10 @@ class SitemapUrlInformationProvider implements UrlInformationProviderInterface
         $urlInformation->setLocation($this->router->generate($document, array(), true));
         $urlInformation->setChangeFrequency($this->defaultChanFrequency);
 
+        if (method_exists($document, 'getLastModification')) {
+            $urlInformation->setLastModification($document->getLastModification());
+        }
+
         if ($this->alternateLocaleProvider) {
             $collection = $this->alternateLocaleProvider->createForContent($document);
             $urlInformation->setAlternateLocales($collection->toArray());


ElectricMaxxx:
Can you have a look into #225 ? Maybe i still did that change there? Cause that PR does a bigger refactoring in the handling creating the values  - not in the usage.

ElectricMaxxx:
Yea and now as i saw your code, i realized that it would implement a very special usecase. And that is the reason why i started to Do #225 cause you will be able to implement and register (tagged Service) to guess/vote the values of `UrlInformation`.

WillBBP:
Hi, I'd like to have a go at making our stuff run with your #255 and see how it goes. Currently getting...

`InvalidConfigurationException: Invalid configuration for path "cmf_seo": Notice: Undefined index: content_listener in /vagrant/vendor/symfony-cmf/seo-bundle/DependencyInjection/Configuration.php line 48`

I'm guessing this isn't enough config anymore...

    cmf_seo:
        persistence:
            phpcr:
                enabled: true
                manager_name: ~
        sonata_admin_extension: true
        sitemap: true

Can you give me some info on the new configuration options?

ElectricMaxxx:
That's a feature @dbu implemented or refectored. Maybe he has got a fast idea.

dbu:
i will try to find some time tomorrow to work on the SeoBundle pull
request and hopefully get it ready to merge.

meanwhile, you might find a hint about the configuration looking at the
diff of #225 - or maybe we just have some bug in there. (well, we
certainly at least have a bug that we should report a clear error
instead of the fatal error you get)

ElectricMaxxx:
@dbu but we shouldn't merge this one, right? (cause it is a detailed implementation) So i would suggest to implement a guesser with that implementation later.

dbu:
yes lets wait until #225 is merged and then refactor this one (if its not already in #225)

dbu:
okay, #225 has been merged. we only just started writing documentation, but the new approach would be to implement the GuesserInterface and tag it as guesser with a higher priority to set last modification before its set by something else. does that make sense? imho it makes more sense to build that in your project than inside this bundle. wdyt?

WillBBP:
If only "business" hadn't just canned everything that was in our sprint and moved us onto something else with a crazy deadline :-(

dbu:
oh noes! hope you can get back to this later.
ok if we close this issue then?

as a potential user of the bundle in a different context, your input in
#222 would be appreciated btw.

WillBBP:
Yes, close it for now

PR #244, opened 2015-07-28
configuring the default templates in configuration directly

dbu:
@ElectricMaxxx i would suggest to do like this instead.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -114,6 +114,10 @@ class Configuration implements ConfigurationInterface
                                 ->arrayNode('templates')
                                     ->useAttributeAsKey('format')
                                     ->requiresAtLeastOneElement()
+                                    ->defaultValue(array(
+                                        'html' => 'CmfSeoBundle:Sitemap:index.html.twig',
+                                        'xml' => 'CmfSeoBundle:Sitemap:index.xml.twig',
+                                    ))
                                     ->prototype('scalar')->end()
                                 ->end()
                             ->end()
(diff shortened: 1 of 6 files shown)


Tests/Unit/DependencyInjection/CmfSeoExtensionTest.php
```diff
@@ -368,8 +368,8 @@ public function testDefaultTemplatesSet()
             array(
                 'sitemap' => array(
                     'templates' => array(
-                        'xml' => CmfSeoBundle::DEFAULT_SITEMAP_TEMPLATE_XML,
-                        'html' => CmfSeoBundle::DEFAULT_SITEMAP_TEMPLATE_HTML,
```
dbu: tests should not rely on such a default value or you won't detect accidental change. and changing that template name would be a BC break for people that overwrote the template rather than configuring one explicitly.

DependencyInjection/Configuration.php
```diff
@@ -114,6 +114,10 @@ public function getConfigTreeBuilder()
                                 ->arrayNode('templates')
                                     ->useAttributeAsKey('format')
                                     ->requiresAtLeastOneElement()
+                                    ->defaultValue(array(
+                                        'html' => 'CmfSeoBundle:Sitemap:index.html.twig',
+                                        'xml' => 'CmfSeoBundle:Sitemap:index.xml.twig',
+                                    ))
```
ElectricMaxxx: did not know how to do defaultValues on aray node, cool. Now i do.

ElectricMaxxx:
thanks.

Issue #242, opened 2015-07-23
option to disable default loader / voter / guesser


dbu:
the default things should be possible to disable. not sure how to not make this horribly verbose... ideas? just a global on/off?

ElectricMaxxx:
Doesn't WE have a priority to overwrite them? This would he the first solution. 
The second would be to create a enable/disable map which can be passed to the voters. They can decline when their own option is set to false. Same for guesser, they can decline in `supports()` method.

dbu:
for guessers yes, only if i want some value not at all, there is still a problem. for the loader, its more important because i might have phpcr enabled but need my own, more clever, loader instead of the default one...

maybe a section in the defaults? 
```yaml
cmf_seo:
    sitemap:
        defaults:
            loaders: none|all|phpcr
            guessers: none|all|<comma separated list>
            voters: none|all|publish-workflow            
```

then in the DI extension we just remove the services from the container that are not activated.

ElectricMaxxx:
Or just remove the tag?

dbu:
why would we keep the services lying around? we actually should even
make them private, no point in other things accessing them.

ElectricMaxxx:
:+1:

ElectricMaxxx:
`none` would be default for all or `all`?

ElectricMaxxx:
seems wrong and hard to implement to have a different usage on them, so i would suggest to use `none|all|<comma-sepparated-list>` for all.

ElectricMaxxx:
What about this option for each sitemap? What about using defaults for each sitemap? What about one sitemap whants the `guesser/loader/voter` the other doesn't? Or do you want it as a global flag?

ElectricMaxxx:
solved by #245#

PR #245, opened 2015-07-28
option to disable guessers/loaders/voters globaly

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #242 
| License       | MIT
| Doc PR        | https://github.com/symfony-cmf/symfony-cmf-docs/issues/693

Started implementing with that and came to an easy global on/off/partial-off solution, we can start with. Doing some fance inheritance into the specific sitemap configurations will be an overkill atm - i think.

When you like that, i will do:
* [x] configuration schema extension
* [x] tests for configuration schema extension

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,17 @@
 Changelog
 =========
 
-* **2015-07-20**: Cleaned up the sitemap generation. If you used the unreleased version of 
-  sitemaps, you will need to adjust your code. See https://github.com/symfony-cmf/SeoBundle/pull/225
-* **2015-02-24**: Configuration for content_key moved to content_listener section, and its now possible to disable
-  The content listener by setting cmf_seo.content_listener.enabled: false
+* **2015-07-20**: Cleaned up the sitemap generation. If you used the unreleased 
+  version of sitemaps, you will need to adjust your code. See https://github.com/symfony-cmf/SeoBundle/pull/225
+  Options are available to keep all or no voters|guessers|loaders enabled or 
+  enable them one by one by their service id.
+* **2015-02-24**: Configuration for `content_key` moved to the `content_listener` 
+  section, and its now possible to disable the content listener by setting 
+  `cmf_seo.content_listener.enabled: false`
 * **2015-02-14**: Added sitemap generation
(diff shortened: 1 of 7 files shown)


dbu:
for sitemap specific, we could have the same config in each configuration, and update the service tags with the sitemap attribute. but yeah, lets leave that out for now.

i would invert `all` and `none`, its confusing right now (the config does not say it would be about which services to skip). saying which to use, with default `all` sounds clearer to me.

i would make the configuration options arrayNode instead of comma separated. we can have a pre-task to convert a single string with "all" or also "...loader.phpcr" to an array.

i wonder if all/none could clash with a sitemap name. maybe use _all / _none? or something else that is very unlikely to be a sitemap? particularly `all` sounds a bit risky.

dbu:
oh, and lets start a separate doc PR so we get the first one merged at some point...

ElectricMaxxx:
@dbu as the issue starts with `disable ... ` i implemented it from the position, which sitemaps to disable :-) but i can change it if you want to.

I started implementing the configuration as an `enumNode`, but had problems to describe the custom list as values :-( So you really want to blow up the `Configuration.php` just to get the list as a pre-task? I think i can devide the values better as strings, the list is just one of the possible options, and i would have to extract `all`/`none` when they are the values.

To set the same option into each sitemap would be not the problem, to disable the services for more then one sitemap, would be very hard. I would imagine to steel the service caused by the option on one sitemap and an other still whants the service ... So lets just keep it global for now.

ElectricMaxxx:
@dbu i don't understand the thing with mixing sitemap names and all|none the options never touch a sitemap by its name. in the worst case a service with the id all|none can't be implemented.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* **2015-07-30**: implement options to enable/disable guessers, loaders and voters
```
dbu: i would just add that to the rest of the sitemap thing

ElectricMaxxx:
@dbu we should merge that, right?

dbu:
i squashed commits. i think its now ready to merge. i created a separate issue for the doc, so that we don't block the initial doc PR any longer.

Issue #240, opened 2015-07-23
is SitemapElementInterface still used?


dbu:
seems to be essentially unused, none of the new sitemap infrastructure is using it.

https://github.com/symfony-cmf/SeoBundle/blob/master/SitemapElementInterface.php

we might add the sitemap name to the method and create a voter for it, if we think we should have that, e.g. for an admin flag "show in sitemap".

ElectricMaxxx:
I think this interface isn't used anymore. The voters decide whether an element is included or not.

dbu:
currently there is one voter, the publish workflow voter. it could make sense to have documents expose a flag whether they want to be on the sitemap. alternatively it could be a method to get the list of sitemaps the document is to be included. but that would need a way to say "all" and "nothing".

ElectricMaxxx:
@dbu so you would implement a simple voter to check that interface and the value of the flag?

dbu:
yes, would make sense with the new architecture i think. and be pretty
trivial, but a valid use case. the dev can then either expose a flag to
the editors, or have some simple logic on a model class to decide.

PR #247, opened 2015-07-30
implement voter for SitemapAwareDocuments

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #240 
| License       | MIT
| Doc PR        |


Sitemap/SitemapAwareDocumentVoter.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Implementing the SitemapAwareInterface will give the chance the let the document decide
+ * whether it wants to be displayed on sitemap or not.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class SitemapAwareDocumentVoter implements VoterInterface
+{
+
```
dbu: extra blank line

Sitemap/SitemapAwareDocumentVoter.php
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function exposeOnSitemap($content, $sitemap)
+    {
+        return $content instanceof SitemapAwareInterface && $content->isVisibleInSitemap();
```
dbu: the voters need to be unanimous, right? and there is no 3rd option "abstain"? i think whe need to say `if (!$content instanceof SitemapAwareInterface) {return true;}`

SitemapAwareInterface.php
```diff
@@ -15,13 +15,9 @@
...
  * Models implementing this interface can indicate whether they should be listed in the sitemap or not.
  *
- * For PHPCR-ODM documents used with the \Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr\SitemapUrlInformationProvider,
- * make sure to call the property that stores this information "visible_for_sitemap",
- * as the provider uses the property in the query rather than loading all documents and checking the method.
- *
  * @author Maximilian Berghoff <Maximilian.Berghoff@gmx.de>
  */
-interface SitemapElementInterface
+interface SitemapAwareInterface
```
dbu: should the method signature accept the sitemap name? would be the consequent thing with the rest of how sitemap is built. the document is still free to ignore the name if its a global flag.

Sitemap/SitemapAwareDocumentVoter.php
dbu: if we don't split that, every content that does not implement the interface will automatically be hidden from all sitemaps. imo if we want only such models, it would be the job of the provider to not load others in the first place.

ElectricMaxxx: i think, we just have a `true`/`false` return on the voters, right?

Sitemap/SitemapAwareDocumentVoter.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+class SitemapAwareDocumentVoter implements VoterInterface
+{
+
+    /**
+     * {@inheritdoc}
+     */
+    public function exposeOnSitemap($content, $sitemap)
+    {
+        return $content instanceof SitemapAwareInterface && $content->isVisibleInSitemap($sitemap);
```
dbu: somehow my comment disapeared. but i still think we need to return true if the content is not instanceof the interface. either with an if, or 
``return !$content instanceof SitemapAwareInterface || $content->isVisibleInSitemap($sitemap);``

Sitemap/SitemapAwareDocumentVoter.php
ElectricMaxxx: ahhh now i understand.

dbu:
lets squash the commits and merge, looks good!

ElectricMaxxx:
no, forgot to change my tests that fits the preor case

ElectricMaxxx:
merge this when happy.

dbu:
this also needs to be documented

ElectricMaxxx:
Wrote https://github.com/symfony-cmf/symfony-cmf-docs/issues/692 for that.

PR #248, opened 2015-07-31
allow to configure sitemap helpers globally and per sitemap

dbu:
that was a bit of a braintwister. wdyt?

* rename "all" to "_all" and "none" to "_none" to avoid name clashes
* move more logic into configuration class
* allow to configure per sitemap


DependencyInjection/CmfSeoExtension.php
```diff
@@ -252,29 +258,48 @@ private function loadSitemapHandling($config, XmlFileLoader $loader, ContainerBu
         $loader->load('sitemap.xml');
 
         $configurations = $config['configurations'];
-        // if there are no explicit configurations, enable the default sitemap
-        if (!count($configurations)) {
-            $configurations['sitemap'] = array();
```
dbu: moved to configuration class so that we get defaults set up by the configuration logic

DependencyInjection/CmfSeoExtension.php
```diff
@@ -252,29 +258,51 @@ private function loadSitemapHandling($config, XmlFileLoader $loader, ContainerBu
...
+
+        $helperStatus = array();
+        foreach ($this->sitemapHelperMap as $helper => $tag) {
+            $helperStatus[$helper] = array();
+            $serviceDefinitionIds = $container->findTaggedServiceIds($tag);
+            foreach ($serviceDefinitionIds as $id => $attributes) {
+                if (0 === strncmp($this->getAlias(), $id, strlen($this->getAlias()))) {
+                    // avoid interfering with services that are not part of this bundle
```
dbu: if we don't do this, behaviour would depend on order in which bundles are instantiated

Tests/Resources/Fixtures/config/config2.xml
```diff
@@ -11,9 +11,12 @@
     <error enable-parent-provider="true" enable-sibling-provider="true" />
 
     <sitemap enabled="true">
-        <configuration default-change-frequency="never" name="default" guessers="none" voters="none" loaders="none">
+        <configuration default-change-frequency="never" name="default">
             <template format="xml">test.xml</template>
             <template format="html">test.html</template>
+            <loader>_none</loader>
+            <voter>_none</voter>
+            <guesser>_none</guesser>
```
dbu: this is more verbose now. @WouterJ what is best practice for the xml schema? should we allow both elements and the attribute that is converted by the configuration class to a list?

WouterJ: I would recommend to choose to use either attributes or elements, in this case (As it's a prototyped array) I would choose elements.

Otherwise, you have to write some conditionals (to not allow elements *and* attributes at the same time), which is not supported by PHP's DomDocument atm IIRC.

Resources/config/schema/seo-1.0.xsd
```diff
@@ -70,12 +70,12 @@
...
             <xsd:element name="template" type="sitemap_template" minOccurs="1" maxOccurs="unbounded" />
+            <xsd:element name="loader" type="xsd:string" minOccurs="0" maxOccurs="unbounded" />
+            <xsd:element name="voter" type="xsd:string" minOccurs="0" maxOccurs="unbounded" />
+            <xsd:element name="guesser" type="xsd:string" minOccurs="0" maxOccurs="unbounded" />
         </xsd:sequence>
         <xsd:attribute name="default-change-frequency" type="change_frequency" default="always" />
 
-        <xsd:attribute name="guessers" type="xsd:string" />
-        <xsd:attribute name="loaders" type="xsd:string" />
-        <xsd:attribute name="voters" type="xsd:string" />
```
ElectricMaxxx: why did you change from attribute to element?

dbu: because its a arrayNode now. i remove the split on `,` because that
seems like a hack to me.

dbu:
@ElectricMaxxx this is green. now you have to decide if you like it or not...

ElectricMaxxx:
what a question, your sollution is awesome.

Issue #201, opened 2014-12-11
Default SeoMetadata class


dbu:
@goemaere found that when using the [conference tutorial](https://github.com/dbu/conference-tutorial/pull/16) and running in production mode, creating a new page and then loading it again fails. after some debugging, i found that this is because the sonata extension [form type sets data_class to the model class](https://github.com/symfony-cmf/SeoBundle/blob/master/Form/Type/SeoMetadataType.php#L64). this does not work out because we store that in the db and get a doctrine error that there is no proxy class available. (in terms of doctrine, the model class is not complete, having no idea - there is probably a missing doctrine sanity check, it should prevent storing this if it does not work)

either we need to document that the user MUST init SeoMetadata to the Doctrine\Phpcr class in the page constructor or we make the default class configurable and inject it to the form type. i prefer the later.

kminh:
Not sure if related to this issue, but using version 1.1 of this bundle, when go into production this file is missing: `app/cache/prod/__CG__SymfonyCmfBundleSeoBundleModelSeoMetadata.php`.

Do you have any work around for now? Currently I have to copy this file from the dev environment.

ElectricMaxxx:
Does look related. Looks like you forgot to clear cache for prod Environment.

kminh:
I did, but that proxy class doesn't get generated when clearing the cache. When accessing a page with SeoMetadata attached in dev environment that proxy class is generated.

dbu:
i think this is exactly the issue. can you check what class is stored in the class property of the node? `app/console doctrine:phpcr:node:dump /path/to/document --properties --depth=0` if it is the Model\SeoMetadata, that is the problem. 

you could run a phpcr-shell query to update the class name to the one in Doctrine\Phpcr.

and to prevent this from happening in the future, you need to create the metadata in the constructor of your documents as the form autocreates it to the model class which will not work out.

kminh:
`app/console doctrine:phpcr:node:dump /cms/content/pages/name/seoMetadata --props --depth=0` returns: `phpcr:class = Symfony\Cmf\Bundle\SeoBundle\Model\SeoMetadata`

By creating the metadata in the constructor, do you mean this:
```php
// in Acme/AcmeBundle/Doctrine/Phpcr/Page
use Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr\SeoMetada;

class Page
{
    public function __constructor()
    {
        $this->seoMetadata = new SeoMetada();
    }
[...]

PR #249, opened 2015-08-10
fix seo metadata data_class in form type

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #201 
| License       | MIT

### DIFF
--- a/DependencyInjection/CmfSeoExtension.php
+++ b/DependencyInjection/CmfSeoExtension.php
@@ -95,6 +95,12 @@ class CmfSeoExtension extends Extension
         if ($this->isConfigEnabled($container, $config['alternate_locale'])) {
             $this->loadAlternateLocaleProvider($config['alternate_locale'], $container);
         }
+
+        $this->loadFormConfiguration(
+            $config['form'],
+            $container,
+            $this->isConfigEnabled($container, $config['persistence']['phpcr']) ? 'phpcr' : 'default'
+        );
     }
 
     /**
@@ -342,5 +348,26 @@ class CmfSeoExtension extends Extension
                 }
             }
         }
+    }
+
+    /**
+     * Configuration block to configure form.
+     *
+     * The data_class option form type for the SeoMetadata depends on the chose storage.
+     *
+     * @param array            $config
+     * @param ContainerBuilder $container
+     * @param string           $storage   Information about the configured storage.
+     */
+    private function loadFormConfiguration($config, ContainerBuilder $container, $storage)
+    {
+        $seoMetadataClass = 'Symfony\Cmf\Bundle\SeoBundle\Model\SeoMetadata';
+        if (null !== $config['data_class']['seo_metadata']) {
+            $seoMetadataClass = $config['data_class']['seo_metadata'];
+        } elseif ('phpcr' === $storage) {
+            $seoMetadataClass = 'Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr\SeoMetadata';
+        }
+
+        $container->setParameter($this->getAlias() . '.form.data_class.seo_metadata', $seoMetadataClass);
     }
 }
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -79,7 +79,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('title')->end()
                 ->scalarNode('description')->end()
                 ->scalarNode('original_route_pattern')->defaultValue(SeoPresentation::ORIGINAL_URL_CANONICAL)->end()
-       ;
+        ;
 
         $this->addPersistenceSection($nodeBuilder);
         $this->addSonataAdminSection($nodeBuilder);
@@ -87,6 +87,9 @@ class Configuration implements ConfigurationInterface
         $this->addErrorHandlerSection($nodeBuilder);
         $this->addSitemapSection($nodeBuilder);
         $this->addContentListenerSection($nodeBuilder);
+        $this->addFormSection($nodeBuilder);
+
+        $nodeBuilder->end();
 
         return $treeBuilder;
     }
@@ -278,5 +281,27 @@ class Configuration implements ConfigurationInterface
                 ->end()
             ->end()
         ;
+    }
+
+    /**
+     * Attach the form node to the tree.
+     *
+     * @param NodeBuilder $nodeBuilder
+     */
+    private function addFormSection($nodeBuilder)
+    {
+        $nodeBuilder
+           ->arrayNode('form')
+                ->addDefaultsIfNotSet()
+                ->children()
+                    ->arrayNode('data_class')
+                        ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('seo_metadata')->defaultNull()->end()
+                        ->end()
+                    ->end()
+                ->end()
+           ->end()
+           ;
     }
 }
--- a/Form/Type/SeoMetadataType.php
+++ b/Form/Type/SeoMetadataType.php
@@ -29,6 +29,19 @@ use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 class SeoMetadataType extends AbstractType
 {
     /**
+     * @var
+     */
+    private $dataClass;
+
+    /**
+     * @param string $dataClass The FQCN of the data class to use for this form.
+     */
+    public function __construct($dataClass)
+    {
+        $this->dataClass = $dataClass;
+    }
+
+    /**
      * {@inheritDoc}
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
@@ -70,7 +83,7 @@ class SeoMetadataType extends AbstractType
     public function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults(array(
-            'data_class' => 'Symfony\Cmf\Bundle\SeoBundle\Model\SeoMetadata',
+            'data_class' => $this->dataClass,
             'translation_domain' => 'CmfSeoBundle',
             'required' => false,
         ));
--- a/Tests/WebTest/Admin/SeoContentAdminExtensionTest.php
+++ b/Tests/WebTest/Admin/SeoContentAdminExtensionTest.php
@@ -12,7 +12,6 @@
 namespace Symfony\Cmf\Bundle\SeoBundle\Tests\WebTest\Admin;
 
 use Symfony\Cmf\Component\Testing\Functional\BaseTestCase;
-use Symfony\Component\HttpKernel\Client;
 
 /**
  * This test will cover all behavior with the provides admin extension.
@@ -68,5 +67,20 @@ class SeoContentAdminExtensionTest extends BaseTestCase
         $this->assertEquals(200, $res->getStatusCode());
         $this->assertCount(1, $crawler->filter('html:contains("Key")'));
         $this->assertCount(1, $crawler->filter('html:contains("Value")'));
+    }
+
+    public function testItemCreate()
+    {
+
+        $crawler = $this->client->request('GET', '/admin/cmf/seo/seoawarecontent/create');
+        $res = $this->client->getResponse();
+
+        $this->assertEquals(200, $res->getStatusCode());
+
+        $this->assertCount(1, $crawler->filter('html:contains("SEO")'));
+        $this->assertCount(1, $crawler->filter('html:contains("Page title")'));
+        $this->assertCount(1, $crawler->filter('html:contains("Original URL")'));
+        $this->assertCount(1, $crawler->filter('html:contains("description")'));
+        $this->assertCount(1, $crawler->filter('html:contains("keywords")'));
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* **2015-08-12**: Added configuration for the default data class of the `seo_metadata` form type.
 * **2015-07-20**: Cleaned up the sitemap generation. If you used the unreleased 
   version of sitemaps, you will need to adjust your code. See https://github.com/symfony-cmf/SeoBundle/pull/225
   Options are available to keep all or no voters|guessers|loaders enabled or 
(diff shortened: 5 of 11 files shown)


Form/Type/SeoMetadataType.php
```diff
@@ -83,4 +95,28 @@ public function getName()
...
         return 'seo_metadata';
     }
+
+    /**
+     * @param SeoMetadata $data
+     * @return null|SeoMetadataPersistenceModel
+     */
+    private function convertToPersisterModel(SeoMetadata $data)
+    {
+        if (!class_exists('Doctrine\Bundle\PHPCRBundle\DependencyInjection\Compiler\DoctrinePhpcrMappingsPass')
```
ElectricMaxxx: I tried to inject `cmf_seo.backend_type_phpcr` and `cmf_seo.backend_type_orm` to recognize the persistence type, but that does noch work, as both parameters arn't available together. :-(
So this was my only chance to do it atm.

Tests/WebTest/Admin/SeoContentAdminExtensionTest.php
```diff
@@ -69,4 +68,21 @@ public function testExtraPropertyEditView()
...
+
+        $this->assertEquals(200, $res->getStatusCode());
+
+        $this->assertCount(1, $crawler->filter('html:contains("SEO")'));
+        $this->assertCount(1, $crawler->filter('html:contains("Page title")'));
+        $this->assertCount(1, $crawler->filter('html:contains("Original URL")'));
+        $this->assertCount(1, $crawler->filter('html:contains("description")'));
+        $this->assertCount(1, $crawler->filter('html:contains("keywords")'));
+
+        // Todo: Type in some data and send them, extract the id and check the database values
```
ElectricMaxxx: this schould be finalized, to see the result.

dbu:
i think the better way to do this would be to [set the default data_class](http://symfony.com/doc/current/book/forms.html):

```
use Symfony\Component\OptionsResolver\OptionsResolver;

public function configureOptions(OptionsResolver $resolver)
{
    $resolver->setDefaults(array(
        'data_class' => 'AppBundle\Entity\Task',
    ));
}
```

Form/Type/SeoMetadataType.php
dbu: we use `cmf_seo.backend_type_phpcr` for the compiler pass. this is set in the DI extension. we can add a constructor parameter for the type that sets the seo metadata class default and add that parameter in the DI extension.

ElectricMaxxx: No we don't really use it, we use the if-statemant for the existence of
the Doctrine classes and then use the parameter. But our Extension class
only set `backend_type_orm` OR `backend_type_phpcr`. So one of them
isn't available. I tried to set the oposite (backend_type_orm = false,
for phpcr), but then the the application was broken again.

ElectricMaxxx:
Yea, i know that parameter, but can we set it based on the persistence
decission?

Form/Type/SeoMetadataType.php
```diff
@@ -69,8 +76,15 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
      */
     public function configureOptions(OptionsResolver $resolver)
     {
+        $class = null;
+        if ($this->container->has('cmf_seo.backend_type_phpcr')) {
```
ElectricMaxxx: @dbu you see the issue with the parameter `cmf_seo.backend_type_phpcr` now? This one ist not available, when `ORM` is used `cmf_seo.backend_type_orm` instead. Both are just set to true, and have no false as default.

dbu: hm. why not have a configuration option for the seo metadata class that defaults to the phpcr document if phpcr is active and to the model otherwise? and have the DI set that parameter on the container, and use constructor injection?

ElectricMaxxx: when having a configuration option or a parameter created in the Extension class, i wouldn't need the container anymore, and yes i would use the constructor injection then.
I think creating a `seo_meta_data_class` parameter based on the persistence decission, would be the easiest way. I wouldn't make it as a configuration, why should we?

dbu: if somebody is using doctrine orm or something else, they might want to
overwrite it. but we can start without making it configurable, it can
still be overwritten when using the form type.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -290,4 +296,29 @@ private function loadSitemapHandling($config, XmlFileLoader $loader, ContainerBu
...
+     *
+     * @param array            $config
+     * @param ContainerBuilder $container
+     */
+    private function loadFormConfiguration($config, ContainerBuilder $container)
+    {
+        $seoMetadataClass = null;
+        if (isset($config['data_class'])
+            && isset($config['data_class']['seo_metadata'])
+            && null !== $config['data_class']['seo_metadata']
```
ElectricMaxxx: I don't really like that. Should i default the complete array for `data_class`?

DependencyInjection/Configuration.php
```diff
@@ -225,4 +227,25 @@ private function addContentListenerSection(NodeBuilder $nodeBuilder)
...
+     *
+     * @param NodeBuilder $nodeBuilder
+     */
+    private function addFormSection($nodeBuilder)
+    {
+        $nodeBuilder
+           ->arrayNode('form')
+                ->canBeDisabled()
+                ->children()
+                    ->arrayNode('data_class')
```
ElectricMaxxx: I think i need to fix xml or yml here, right?

ElectricMaxxx: or not. We have both `_` and `-`. We schould fix that for a version 2.0

dbu: i think there is no need to fix xml config. that is done when you have a list of entries that in yml have a group name with an "s" at the end but in xml you can just have a list of elements with the name. as in: `sitemaps: {yml array}` vs `<sitemap>...</sitemap><sitemap></sitemap>`

DependencyInjection/Configuration.php
```diff
@@ -225,4 +227,25 @@ private function addContentListenerSection(NodeBuilder $nodeBuilder)
...
+    /**
+     * Attach the form node to the tree.
+     *
+     * @param NodeBuilder $nodeBuilder
+     */
+    private function addFormSection($nodeBuilder)
+    {
+        $nodeBuilder
+           ->arrayNode('form')
+                ->canBeDisabled()
```
dbu: do we need to allow to disable this? i would not see the need for an enabled field.

dbu: and i think if you do addDefaultsIfNotSet you will not need the checks on data_class in the DI class.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -90,6 +92,10 @@ public function load(array $configs, ContainerBuilder $container)
         if ($this->isConfigEnabled($container, $config['alternate_locale'])) {
             $this->loadAlternateLocaleProvider($config['alternate_locale'], $container);
         }
+
+        if ($this->isConfigEnabled($container, $config['form'])) {
```
dbu: if we drop the canBeDisabled this could go and we would just load unconditionally.

DependencyInjection/Configuration.php
ElectricMaxxx: had some complications with the existence of the the embedded array and so on, will try to do that better.

DependencyInjection/CmfSeoExtension.php
dbu: or just `if (!empty($config['data_class']['seo_metadata']))` . !empty is both isset and not null.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -290,4 +296,29 @@ private function loadSitemapHandling($config, XmlFileLoader $loader, ContainerBu
...
+    {
+        $seoMetadataClass = null;
+        if (isset($config['data_class'])
+            && isset($config['data_class']['seo_metadata'])
+            && null !== $config['data_class']['seo_metadata']
+        ) {
+            $seoMetadataClass = $config['data_class']['seo_metadata'];
+        } else {
+            if ($container->hasParameter($this->getAlias() . '.backend_type_phpcr')) {
+                $seoMetadataClass = self::CLASS_SEO_METADATA_PHPCR;
```
dbu: as we only use this once, i don't think we need a constant.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -290,4 +296,29 @@ private function loadSitemapHandling($config, XmlFileLoader $loader, ContainerBu
...
+    private function loadFormConfiguration($config, ContainerBuilder $container)
+    {
+        $seoMetadataClass = null;
+        if (isset($config['data_class'])
+            && isset($config['data_class']['seo_metadata'])
+            && null !== $config['data_class']['seo_metadata']
+        ) {
+            $seoMetadataClass = $config['data_class']['seo_metadata'];
+        } else {
+            if ($container->hasParameter($this->getAlias() . '.backend_type_phpcr')) {
```
dbu: this is a bit weird. i would have a parameter `$storage` to this method which you pass as `$this->isConfigEnabled($container, $config['persistence']['phpcr'] ? 'phpcr' : 'default'`. then you can determine the class based on the explicit argument.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -27,6 +26,9 @@
  */
 class CmfSeoExtension extends Extension
 {
+    const CLASS_SEO_METADATA_PHPCR = 'Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr\SeoMetadata';
+    const CLASS_SEO_METADATA_ORM = 'Symfony\Cmf\Bundle\SeoBundle\Document\SeoMetadata';
```
dbu: the namespace is Model, not Document. and i would call it the "default" class, not the "orm" class as its a model class. but as said below, i would not use constants in this case.

DependencyInjection/Configuration.php
dbu: afaik symfony fixes that automatically, when we use _ in our config that
you can use - in xml.

DependencyInjection/CmfSeoExtension.php
ElectricMaxxx: oha, what a fault.

DependencyInjection/CmfSeoExtension.php
ElectricMaxxx: :+1:

ElectricMaxxx: but why do we have that parameter then? Just for the MappingPath inside the Bundle class?

dbu: yes, its for the mapping pass to know if it actually should map. this is because we need to register compiler passes before we know the configuration, so they run unconditionally and hae to check themselves if they are needed.

DependencyInjection/CmfSeoExtension.php
dbu: if the build was green, this indicates a missing test :-)

ElectricMaxxx: Yea the Frontend ORM Test for the AdminView.

ElectricMaxxx: And ... the persistence of both (PHPCR and ORM) by the AdminView, we
don't have both. Cause i think the effect will be visible on clicking
the "Save" Button only, right?

dbu: i think even just loading the form would fail as it would try to create
the data_class. not sure though.

ElectricMaxxx: So i have to create the admin views (edit, create) with orm content, but
we don't have AdminTests for ORM at all :-(

ElectricMaxxx: And i Know the reason why:
We are able to load and do unit tests based on the persistence setting,
but we have no chance with the current version of the TestinBundle to
switch configuration for FrontendTest. We have on configuration loaded
and thats the PHPCR one.

dbu: well, we can say we do that "later"

ElectricMaxxx: thanks. :-)

ElectricMaxxx:
So it is green, now. It the second to last bug before tagging the seo bundle. lets merge it and put a hint into the documentation.

ElectricMaxxx:
unfortuneately i merged (we do it in the current project).

DependencyInjection/CmfSeoExtension.php
```diff
@@ -343,4 +349,23 @@ private function handleSitemapHelper($helperStatus, ContainerBuilder $container)
             }
         }
     }
+    
+    /**
+     * Configuration block to configure form.
+     *
+     * @param array            $config
+     * @param ContainerBuilder $container
+     * @param string           $storage
```
dbu: maybe describe what this parameter is?

DependencyInjection/Configuration.php
```diff
@@ -78,8 +78,9 @@ public function getConfigTreeBuilder()
                 ->scalarNode('translation_domain')->defaultValue('messages')->end()
                 ->scalarNode('title')->end()
                 ->scalarNode('description')->end()
-                ->scalarNode('original_route_pattern')->defaultValue(SeoPresentation::ORIGINAL_URL_CANONICAL)->end()
-       ;
+                ->scalarNode('original_route_pattern')->defaultValue(SeoPresentation::ORIGINAL_URL_CANONICAL)
+            ->end()
```
dbu: the `;` should be on a new line, but the end you moved down ends the original_route_pattern node. correct would be to add another end()

ElectricMaxxx: ... and again a error hits me while fixing a CS bug on the fly :-)

dbu:
i think this is coming good! i only have cosmetic changes now.

when you squash the commits, the merge will likely go away. and if not i am not religious about git history... :-)

can you add the documentation for this as well? should be a small thing but if we don't do it, the small things acumulate ;-)

ElectricMaxxx:
sqaushed and got a change in the `CHANGELOG.md` now?

dbu:
okay, looks good! thanks a lot!
 can you do the doc PR and we merge this?

dbu:
thanks. created https://github.com/symfony-cmf/symfony-cmf-docs/issues/698 for the doc. can you do that as well? i would just mention it where we mention the form type (that things work automatically from version 1.2 on)

Issue #250, opened 2015-08-10
Exception in Admin use the ErrorHandling for the Frontend


ElectricMaxxx:
IS:
When producing an exception in the Sonata Admin Backend the new `SuggestionProviderController` is called  and tries to serve good matches for errors with http status code 404. And the Twig Error handler uses our template to show the errors. That causes problems with parameters, which arn't set, and we loose information the normal template serves.

SHOULD:
The `SuggestionProviderController` should get an exclusion list (pattern) for urls, that it does not have to work on. We can't override the main twig-Template, too. So we should create and use an own one, when the `SuggestionProviderController` really has to work.

dbu:
now only documentation for this is missing, and maybe setup in the sandbox

PR #251, opened 2015-08-17
BUG #250 - fix errror handling

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #250
| License       | MIT
| Doc PR        | https://github.com/symfony-cmf/symfony-cmf-docs/issues/606

This is the first part of the bug:

* [x] use configured templates instead of overriding the twig templates
* [x] exclusion configuration for routes, that should not be covered by the `SuggestionController`

I am struggeling with my self if implementing the exclusion strategy is a bug fix or a new feature. Now it is done :-)


ElectricMaxxx:
Somebody time for a review? Last bug before Tagging. 
Ping @dbu @WouterJ

dbu:
i will try to have a look tonight or tomorrow morning

Controller/SuggestionProviderController.php
```diff
@@ -34,6 +35,21 @@ class SuggestionProviderController extends ExceptionController
     protected $suggestionProviders = array();
 
     /**
+     * A set of configuration.
+     *
+     * @var array
+     */
+    protected $configuration;
```
dbu: private

Controller/SuggestionProviderController.php
```diff
@@ -34,6 +35,21 @@ class SuggestionProviderController extends ExceptionController
...
+
+    /**
+     * @param array $configuration
+     */
+    public function setConfiguration($configuration)
```
dbu: why no constructor injection?

Controller/SuggestionProviderController.php
```diff
@@ -86,4 +103,38 @@ public function addSuggestionProvider(SuggestionProviderInterface $matcher, $gro
...
+     * You can define your templates for each format in the bundle's configuration.
+     *
+     * @param string $format
+     *
+     * @return string
+     */
+    private function getTemplateForSuggestions($format = 'html')
+    {
+        if (!isset($this->configuration['templates'][$format])) {
+            throw new InvalidConfigurationException(
```
dbu: should we not revert to the parent in this case? or return false and let the calling method go to parent::showAction even?

Controller/SuggestionProviderController.php
```diff
@@ -86,4 +103,38 @@ public function addSuggestionProvider(SuggestionProviderInterface $matcher, $gro
...
+                sprintf(
+                    'No template for format %s defined. Use one of %s or configure one for your format.',
```
dbu: i don't think that "Use one of %s" makes sense. I would say "Only the formats %s are defined"

Controller/SuggestionProviderController.php
```diff
@@ -86,4 +103,38 @@ public function addSuggestionProvider(SuggestionProviderInterface $matcher, $gro
...
+
+        return $this->configuration['templates'][$format];
+    }
+
+    /**
+     * @return bool
+     */
+    private function isExcluded(Request $request)
+    {
+        foreach ($this->configuration['excluded_paths'] as $path) {
```
dbu: you could use the request matcher for this. that is a class in symfony that can check a request on path but also controller name and whatnot.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -251,6 +251,12 @@ private function loadErrorHandling($config, ContainerBuilder $container)
                 $container->removeDefinition('cmf_seo.error.suggestion_provider.'.$group);
             }
         }
+
+        $configurations = array();
+        $configurations['templates'] = isset($config['templates']) ? $config['templates'] : array();
+        $configurations['excluded_paths'] = isset($config['excluded_paths']) ? $config['excluded_paths'] : array();
```
dbu: i would use a RequestMatcher and therefore have `excludes` which is a whole hashmap of filter criteria

dbu:
looks good, i have two points: 

* how agressive do we want to explode when no template is configured? no fallback to the default error template system?
* we should use the RequestMatcher to decide whether we want to exclude a request. we did that in FOSHttpCacheBundle and its pretty nice. configuration segment is https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/DependencyInjection/Configuration.php#L245 and we set it up like https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/DependencyInjection/FOSHttpCacheExtension.php#L210-L221 and use it https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/EventListener/AbstractRuleSubscriber.php

Controller/SuggestionProviderController.php
ElectricMaxxx: good point but would be a silent fail and the dev does not know why his template/formate does not work.

ElectricMaxxx: but i think, a well formated hint "when the normal handler comes, you did it wrong" woult be enough

ElectricMaxxx:
Good point with that request matchers will have a look tomorrow morning in the train.

Controller/SuggestionProviderController.php
dbu: > but i think, a well formated hint "when the normal handler comes, you
> did it wrong" woult be enough

:+1:

dbu:
one option could be to just have the configuration in format that allows
to add more than path later and switch to RequestMatcher, but only have
path for now (with the same matching semantics as RequestMatcher of course)

ElectricMaxxx:
implemented the configuration for the matchers now, will do the matcher(s) later, but isn't that code duplication? shouldn't have symfony on service for those matches when parsing routes?

DependencyInjection/Configuration.php
```diff
@@ -179,9 +179,37 @@ private function addErrorHandlerSection(NodeBuilder $nodeBuilder)
     {
         $nodeBuilder
             ->arrayNode('error')
+                ->fixXmlConfig('template')
+                ->fixXmlConfig('excluded_path')
```
WouterJ: this one does not exists?

DependencyInjection/Configuration.php
```diff
@@ -179,9 +179,37 @@ private function addErrorHandlerSection(NodeBuilder $nodeBuilder)
...
+                    ->end()
+                    ->arrayNode('exclusion_rules')
+                        ->info('Rules to exclude error handling from specific matches.')
+                        ->prototype('array')
+                            ->children()
+                                ->arrayNode('match')
+                                    ->children()
+                                        ->scalarNode('path')->defaultNull()->info('Path to exclude')->end()
+                                        ->scalarNode('host')->defaultNull()->info('Host to exclude')->end()
+                                        ->scalarNode('method')->defaultNull()->info('method to exclude')->end()
```
WouterJ: is METHOD really needed in this case? Do we ever want to exclude a method?

Tests/Resources/Fixtures/config/config3.xml
```diff
@@ -0,0 +1,6 @@
+<config xmlns="http://cmf.symfony.com/schema/dic/seo">
+    <error enable-parent-provider="true" enable-sibling-provider="true">
+        <exclusion_rule path="some/path" ip="IP" method="GET" host="test.de" />
+        <exclusion_rule path="some-other/path" ip="IPs" method="POST" host="test-dev.de" />
```
WouterJ: always use dashes instead of underscores in XML

DependencyInjection/Configuration.php
dbu: the idea is to use the symfony RequestMatcher. its probably just
confusing if we don't have all options of the matcher.

DependencyInjection/Configuration.php
ElectricMaxxx: oh yes, you are right should be `exclusion_rule`

Tests/Resources/Fixtures/config/config3.xml
ElectricMaxxx: :+1:

WouterJ:
Hmm, I always prefer whitelisting over blacklisting. However, I can see why this is not really usefull in this case. Still, something in me thinks using exclusion here is wrong.

dbu:
i think in this case blacklisting makes more sense. we could of course do both, but then it can get even more confusing for the user. we could add inclusion_rules the day a use case for them pops up?

regarding duplication: in FOSHttpCacheBundle we defined a base service with the matcher and had our di class extend that for each case. there is not that much duplication except for the setup in the di class, but i don't see how that could be avoided.

DependencyInjection/Configuration.php
WouterJ: Users don't know the implementation behind it. I would prefer to have clear settings, thinking about what is usefull and what not.

E.g., it might be usefull to have a new content type option here. To avoid JSON requests to get the special error page for instance.

ElectricMaxxx:
yesss. i have got the request matcher now. Had some problems to understand the way the services are created on demand by those `DefintionDecorators`, but i works now. 
Will add some doc blocks and better tests tomorrow in the morning. (Need a better separation between default error handling and our new one, and refine the controller tests)

Controller/SuggestionProviderController.php
```diff
@@ -34,6 +35,36 @@ class SuggestionProviderController extends ExceptionController
...
+    /**
+     * @var RequestMatcherInterface
+     */
+    private $exclusionRequestMatcher;
+
+    /**
+     * @param \Twig_Environment       $twig
+     * @param bool                    $debug
+     * @param RequestMatcherInterface $requestMatcher
+     * @param array                   $configuration
```
dbu: maybe explain the requestMatcher and configuration parameter a bit?

Controller/SuggestionProviderController.php
```diff
@@ -34,6 +35,36 @@ class SuggestionProviderController extends ExceptionController
     protected $suggestionProviders = array();
 
     /**
+     * A set of configuration.
+     *
+     * @var array
+     */
+    private $configuration;
```
dbu: can you say what keys and values of this array are?

Controller/SuggestionProviderController.php
```diff
@@ -47,8 +78,9 @@ public function showAction(
         $_format = 'html'
     ) {
         $code = $exception->getStatusCode();
+        $templateForSuggestion = $this->getTemplateForSuggestions($_format);
```
dbu: i would move this to after the "404 || request matcher" check and have a separate return. we should not do the lookup if we do not handle this anyways.

Resources/config/services.xml
```diff
@@ -46,7 +46,17 @@
...
             <argument>%kernel.debug%</argument>
+            <argument type="service" id="cmf_seo.error.exclusion_matcher" />
+            <argument>%cmf_seo.error.configurations%</argument>
         </service>
+
+        <service
+                id="cmf_seo.error.request_matcher"
+                class="Symfony\Component\HttpFoundation\RequestMatcher"
+                public="false"
+                />
```
dbu: indention is off here

dbu:
the build error looks like your configuration does not have the `match` key you ask for in the Configuration class. either adjust the Configuration class or the test fixtures. i think i would remove the match thing. the xml schema also does not have it as far as i see.

ElectricMaxxx:
So how would i define the prototyped array without a key?

ElectricMaxxx:
Got the following error everywher now:

```
[PHPCR\RepositoryException]                                                                                                                              
  Unexpected error talking to the backend: An exception occurred while executing 'SELECT 1 FROM phpcr_workspaces WHERE name = ?' with params ["default"]:  
  SQLSTATE[HY000]: General error: 1 no such table: phpcr_workspaces

````

**Comment by WouterJ** [comment]
You need to bump the dependency on `symfony-cmf/testing` to ^1.3 (or ~1.3)

**Comment by ElectricMaxxx** [comment]
yes looks fine now?

File: Controller/SuggestionProviderController.php [hunk 14]
```diff
@@ -34,6 +35,49 @@ class SuggestionProviderController extends ExceptionController
...
     /**
+     * A set of configuration for the error handling.
+     *
+     * Contains the list of templates defined in the error section of the
+     * cmf_seo configuration.
+     *
+     * Looks like
+     *
+     * array(
+     *  'templates' => array(
```
dbu: why do we inject this nested array and not just the map of templates?

DependencyInjection/Configuration.php
```diff
@@ -179,9 +179,33 @@ private function addErrorHandlerSection(NodeBuilder $nodeBuilder)
...
                 ->children()
                     ->scalarNode('enable_parent_provider')->defaultValue(false)->end()
                     ->scalarNode('enable_sibling_provider')->defaultValue(false)->end()
+                    ->arrayNode('templates')
+                        ->useAttributeAsKey('format')
+                        ->requiresAtLeastOneElement()
+                        ->defaultValue(array(
+                            'html' => 'CmfSeoBundle:Exception:exception.html.twig',
+                            'xml'  => 'TwigBundle:Exception:error.xml.twig',
+                            'json' => 'TwigBundle:Exception:error.json.twig',
```
dbu: should we remove xml and json here and just fall through to the default error controller, now that we fall through on not found templates?

Matcher/ExclusionMatcher.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        $this->matchersMap[] = $ruleMatcher;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function matches(Request $request)
+    {
+        foreach ($this->matchersMap as $element) {
```
dbu: rename $element to $matcher

Controller/SuggestionProviderController.php
ElectricMaxxx: cause i thought there will be more than one configuration later. I think
in Kanban its called `muhda`.

Resources/config/services.xml
```diff
@@ -46,7 +46,16 @@
         <service id="cmf_seo.error.suggestion_provider.controller" class="%cmf_seo.error.suggestion_provider.controller.class%">
             <argument type="service" id="twig" />
             <argument>%kernel.debug%</argument>
+            <argument type="service" id="cmf_seo.error.exclusion_matcher" />
+            <argument>%cmf_seo.error.configurations%</argument>
         </service>
+
+        <service id="cmf_seo.error.request_matcher"
+                 class="Symfony\Component\HttpFoundation\RequestMatcher"
+                 public="false"
```
dbu: should this be abstract=true?

ElectricMaxxx: just copied you own example ;-), but it can.

Controller/SuggestionProviderController.php
dbu: i think i would call it leaking implementation details :-)
but i have no hard opinion on this one, just the feeling that more explicit parameters would be nicer.

Resources/config/services.xml
dbu: you probably would need to un-abstract this with the decorator however, not exactly sure how that works

dbu: if it works i will fix it in FOSHttpCache as well :-)

Controller/SuggestionProviderController.php
WouterJ: :+1: for more explicit parameters

Resources/config/services.xml
WouterJ: decorator automatically unabstracts it.

public=false should also be kept btw

PR #254, opened 2015-08-26
Add tests for lower dependencies

WouterJ:
For some reason, this bundle didn't have lowest deps tests. Let's see what we need to do here

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_install:
   - if [ "$TRAVIS_PHP_VERSION" != "hhvm" ]; then echo "memory_limit = -1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; fi
   - composer require symfony/symfony:${SYMFONY_VERSION} --no-update
   
-instal: composer update $COMPOSER_FLAGS --prefer-dist
+install: composer update $COMPOSER_FLAGS --prefer-dist
 
 before_script: vendor/symfony-cmf/testing/bin/travis/phpcr_odm_doctrine_dbal.sh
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -25,6 +25,8 @@ matrix:
       env: SYMFONY_VERSION=2.8.*
     - php: 5.6
       env: SYMFONY_VERSION=3.0.*
+    - php: 5.3
+      env: SYMFONY_VERSION=2.3.* COMPOSER_FLAGS="--prefer-lowest"
```
ElectricMaxxx: Fails here with: 

``
PHP Fatal error:  Class 'Matthias\SymfonyDependencyInjectionTest\PhpUnit\AbstractCompilerPassTestCase' not found in /home/travis/build/symfony-cmf/SeoBundle/Tests/Unit/DependencyInjection/Compiler/RegisterExtractorsPassTest.php on line 21
```

WouterJ: Yeah, as expect to see more issues because the lower version constraints aren't valid (as they were never tested before this PR).

ElectricMaxxx: you have got a short story for me what that parameter does?

WouterJ: Instead of picking the latest stable version within the version constraint, it'll pick the oldest version possible.

So if we have this constraint for instance: `symfony/symfony: ~2.3`. Normally, this will install 2.7 (or 2.8 if minimum stability is dev). With `--prefer-lowest`, it'll install 2.3.0.

WouterJ:
Did some really ugly things, but tests are now passing.

The ugly bit is that we need to require exact the same PHPunit version to exists in vendor/ as is globally installed on Travis. I've fixed this by parsing the `phpunit --version` output and using that version in the `composer require phpunit/phpunit` command. Not nice, but it's working and it will always work.

WouterJ:
This bundle has a random failing tests. The build failure is unrelated to the changes made in this PR

PR #259, opened 2015-09-02
Fix the dir/baseDir property setting in the FileCache file

coudenysj:
Change the order of the directory creation and property setting so the property holds the correct value

### DIFF
--- a/Cache/FileCache.php
+++ b/Cache/FileCache.php
@@ -35,14 +35,13 @@ class FileCache implements CacheInterface, CacheWarmerInterface, CacheClearerInt
             throw new \InvalidArgumentException(sprintf('The directory "%s" is not writable.', $baseDir));
         }
 
-        $dir = $baseDir.DIRECTORY_SEPARATOR.$dir;
-        if (!is_dir($dir)) {
-            mkdir($dir, 0777 & ~$this->umask, true);
-        }
-
-        $this->dir = rtrim($dir, '\\/');
+        $this->dir = $baseDir.DIRECTORY_SEPARATOR.rtrim($dir, '\\/');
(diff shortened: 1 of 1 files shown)


ElectricMaxxx:
@lsmith77 that seems to be a fix to your fix, right? And i think it should be done on 1.1 and master, right?

Cache/FileCache.php
```diff
@@ -29,20 +29,20 @@ class FileCache implements CacheInterface, CacheWarmerInterface, CacheClearerInt
...
+            throw new \InvalidArgumentException(sprintf('The directory "%s" is not writable.', $baseDir));
         }
 
-        $this->dir = rtrim($dir, '\\/');
-
-        $this->umask = $umask;
-
         $dir = $baseDir.DIRECTORY_SEPARATOR.$dir;
         if (!is_dir($dir)) {
             mkdir($dir, 0777 & ~$this->umask, true);
```
ElectricMaxxx: `$this->unmask` does not exist at, now.

ElectricMaxxx: I think you should use `$unmask`instad, or keep the order for the `$this->unmask = $unmaks`

coudenysj: Correct, let me fine tune a bit

lsmith77:
its not really a fix to my fix .. it just fixes another issue in the same code block .. but yes this should also be put into 1.1

coudenysj:
It fixes the changed line order in 3b8b21c795982df4895bda4896003340a1226a8b.

lsmith77:
ah ok ..

BTW it would be good to add a test for this ..

PR #261, opened 2015-09-06
Merge 1.1

ElectricMaxxx:
Two of the commits should be still in master, just added some test in the PR on 1.1.


ElectricMaxxx:
I am a little bit confused as both existing commits do not cause conflicts, but show the change again.

Issue #146, opened 2014-04-16
Allow usage of cmf bundle for non-PHPCR/cmf documents [RFC]


benglass:
With #145 being addressed I'd like to discuss ways to make this bundle usable for any kind of dynamic seo content management. It could be argued that this out of scope for this bundle but I dont think it would represent significant changes.

Right now there is a listener which looks for the cmf document in the request and updates the metadata based on it if it implements SeoContentAwareInterface.

A listener which goes through all of the attributes to find SeoContentAware items (that were converted by the paramconverter) is probably a bad idea due to the overhead it introduces to request.

I am thinking we could have an annotation that allows you to tag a parameter to a controller as an SeoContentObject. This would then hook into all the existing code for updating the seo presentation/sonata page. This would give you 2 options for having seo content populated by a dynamic object that implements SeoContentAware

1. Pull the seo presentation service out of the container and manually call updateSeoMetadata($object) in your controller action
2. Add an annotation that marks one of your arguments as a SeoContentObject or as SeoContentAware

```php
class FooController
{
    /*
     * @ParamConverter("foo")
     * @SeoContent("foo")
     */
    public function showAction(Foo $foo)
    {
        ...
    }
}
```

Thoughts?

I think having this bundle as a general solution to dynamic metadata management for symfony would be great as i dont think there is a bundle that addresses this (sonata does this for rendering meta tags but is missing the re-usable implementation for extracting seo metadata that this bundle provides). This kind of re-usability could make it a gateway of sorts to the other cmf packages.

dbu:
ping @benglass

benglass:
@dbu Ill take a look at putting together a PR with a twig function for setting seo metadata

benglass:
@dbu is it the CMF coding standard/best practice to always create a helper service that is used for the twig extension? In this case I believe we will only have a single twig method of cmf_seo_update_seo_page but should I still put it inside a service injected into the twig extension a la media bundle?

https://github.com/symfony-cmf/MediaBundle/blob/master/Twig/Extension/CmfMediaExtension.php

benglass:
[...]

PR #263, opened 2015-09-22
Add twig extension to update seo page

benglass:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | ??? (117 deprecation warnings but tests pass)
| Fixed tickets | #146 
| License       | MIT
| Doc PR        | https://github.com/symfony-cmf/symfony-cmf-docs/pull/705

- [X] Pull requests for docs
- [x] Determine if extension dependency on seo presentation should be moved into a SeoHelper
- [ ] Tests?

### DIFF
--- a/Twig/Extension/CmfSeoExtension.php
+++ b/Twig/Extension/CmfSeoExtension.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2015 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @author Ben Glassman <bglassman@gmail.com>
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Twig\Extension;
+
+use Symfony\Cmf\Bundle\SeoBundle\SeoPresentationInterface;
+
+class CmfSeoExtension extends \Twig_Extension
+{
+    /**
+     * @var SeoPresentationInterface
+     */
+    protected $seoPresentation;
+
+    /**
+     * @param SeoPresentationInterface $seoPresentation
+     */
+    public function __construct(SeoPresentationInterface $seoPresentation)
+    {
+        $this->seoPresentation = $seoPresentation;
+    }
+
+    /**
+     * @{inheritdoc}
+     */
+    public function getFunctions()
+    {
+        return array(
+            new \Twig_SimpleFunction('cmf_seo_update_metadata',
+                array($this, 'updateMetadata')
+            ),
+        );
+    }
+
+    /**
+     * @see SeoPresentationInterface
+     */
+    public function updateMetadata($content)
+    {
+        $this->seoPresentation->updateSeoPage($content);
+    }
+
+    public function getName()
+    {
+        return 'cmf_seo';
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* **2015-09-30**: Add `cmf_seo_update_metadata` twig function for updating seo metadata from templates using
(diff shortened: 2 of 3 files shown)


Twig/Extension/CmfSeoExtension.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->seoPresentation = $seoPresentation;
+    }
+
+    /**
+     * @{inheritdoc}
+     */
+    public function getFunctions()
+    {
+        return array(
+            new \Twig_SimpleFunction('cmf_seo_update_seo_page',
```
dbu: why do we call this "update seo page"? i saw the php method is called that way, but the naming seems confused.

* is it not rather a set?
* and why seo *page*?

what about setSeoInformationFromContent($content)? and naming the twig method accordingly? we would need to keep the old php method for BC of course.

benglass: I agree one the name, it seems to be a case where the fact that sonata seo bundle is being used is being exposed perhaps unnecessarily. I think update rather than set may be more correct, however, because some of the data is merged rather than overriden? I'm not 100% sure on this. I think you are saying this but if we are going to change the name here we should probably also change the SeoPresentationInterface method as well and deprecate the old one.

benglass: Perhaps updateSeoMetadataFromContent($content) ?

benglass: @dbu not sure how you are want to proceed with this. I'm happy to rename the public method in the PHP class and twig extension and add a deprecated comment, dunno if anyone else wants to weigh in on what the new name should be. I'd lean towards Metadata vs. information for consistency but I can see the case for set or update.

dbu: sorry for not being responsive, i was away. i created #264 as its an interface change.

for the twig function, could we call it `cmf_seo_update_from_document`? the name you currently have looks weird with the repeated seo. and i rather have a good name here than exact naming consistency with something we agree is not well named.

@ElectricMaxxx wdyt?

benglass: I'm not sure about the use of the word document here as it seems PHPCR specific and the object does not need to be a document, it might be a BlogPost from DoctrineORM or not even any kind of persisted object at all.

dbu: ups, you are correct. `cmf_seo_update_from_model` then? (i guess i know why we ended up with "page" before, but i think page is weird. its probably because in sonata they where thinking about pages from SonataPageBundle)

dbu: so this would be `cmf_seo_update_from`. a bit weird. maybe `cmf_seo_update_metadata`?

benglass: Should we just use the same name for PresentationInterface and this, your recommended updateMetadata

SeoPresentation::updateMetadata($content)
cmf_seo_update_metadata($content)

dbu: yes, thats it. lets go with that. can you update this PR? the work on 2.0 has not yet started, so #264 does not have to happen soon.

benglass: Renamed to cmf_seo_update_metadata and docs PR updated

dbu:
this looks good, thanks! can you please add a note to the CHANGELOG.md file and squash the commits? if you want to add a test, go for it but i can merge without it.

dbu:
the deprecation warnings are because of dependencies like sonata.

benglass:
@dbu Commits squash, changelog updated. I don't think I'm likely to get tests together for this any time soon and it doesnt really seem like its doing very much (just proxying to another method).

dbu:
thanks ben!

PR #270, opened 2016-02-04
[FIX] SeoMetadata.php method names

damonsson:
Change methods to correct `get_class` and `gettype`


dbu:
ups, looks like we lack some unit test here... thanks for the fix!

PR #273, opened 2016-02-09
Fix XSD and a simple deprecation notice

WouterJ:
This feature was implemented to default to the Model class. However, that would never be the correct value. Either the PHPCR document has to be used or one of the custom documents created by the user for their persistence needs.

As you will have hard to debug errors when this setting has a wrong value, I propose to throw an exception instead of defaulting to some faulty value.

Meanwhile, I've fixed the XSD.

/cc @ElectricMaxxx

### DIFF
--- a/Doctrine/Phpcr/AlternateLocaleProvider.php
+++ b/Doctrine/Phpcr/AlternateLocaleProvider.php
@@ -79,7 +79,7 @@ class AlternateLocaleProvider implements AlternateLocaleProviderInterface
 
             $alternateLocaleCollection->add(
                 new AlternateLocale(
-                    $this->urlGenerator->generate($content, array('_locale' => $locale), true),
+                    $this->urlGenerator->generate($content, array('_locale' => $locale), UrlGeneratorInterface::ABSOLUTE_URL),
                     $locale
                 )
             );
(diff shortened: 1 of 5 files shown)


ElectricMaxxx:
Is that related to some kind of open issue?
What about breaking existing applications? Those who didn't set that configuration value will get an exception now, right?

WouterJ:
@ElectricMaxxx this feature was introduced in 1.2-RC1, so it's released just a couple days ago. This is not part of an open issue, I think it's a more user friendly way of implementing the feature. (see https://github.com/symfony-cmf/SeoBundle/pull/249 for the feature PR)

Also, I don't think many people using the SeoBundle use it without the PHPCR persistence layer. In that case, nothing has changed.

ElectricMaxxx:
Seems as you exactly hit your own exception in the tests :-)

WouterJ:
Updated the PR to just do the XSD fixing and one deprecation notice. As ORM uses the model class, defaulting to Model isn't as bad as I thought :see_no_evil:

dbu:
tests complain that some xml does not match the schema.

WouterJ:
Fixed

dbu:
:+1:

PR #274, opened 2016-02-10
Sync error page with core

WouterJ:
* Use a template similar to the one provided by the TwigBundle, only with the suggested pages added
 * Rename the template from 'exception' to 'error', in core exception templates are used in dev, while error templates are used in prod.
 * Fix some config things


Resources/views/Exception/error.html.twig
```diff
@@ -0,0 +1,29 @@
...
+        </p>
+
+        {% if best_matches not is empty %}
+            <h3>Suggested pages</h3>
+
+            {% for group, matches in best_matches %}
+            <h4>{{ group|capitalize }}</h4>
+            <ul>
+                {% for match in matches %}
+                <li><a href="{{ path(match.name) }}">{{ match.name }}</a></li>
```
WouterJ: also added a link. Only listing the titles of relevant pages does not make much sense imo

ElectricMaxxx: :+1:

ElectricMaxxx:
Thanks for the allignment.

PR #277, opened 2016-03-31
Fix form for orm

### LINKED ISSUE #275: Don't persist seoMetadata on Sonata Admin update
Hi,

I'm trying to integrate **SeoBundle 1.1.2** in my project with **Symfony 2.7.9**. and **SonataAdminBundle 2.3.7.**
I use **Doctrine ORM**, I have implemented in my Entity correctly seoMetaData object.

When I create a new record, the information in the fields of seoMetaData, saved correctly, but when editing this record, the values of these fields are **not saved to the database in the update**.

Do you have any idea what could be the problem?

Thank you so much for everything.

dbu:
fix #275

### DIFF
--- a/DependencyInjection/CmfSeoExtension.php
+++ b/DependencyInjection/CmfSeoExtension.php
@@ -78,6 +78,11 @@ class CmfSeoExtension extends Extension
             $sonataBundles[] = 'SonataDoctrineORMBundle';
         }
 
+        $container->setParameter($this->getAlias().'.form_mode_orm',
+            $this->isConfigEnabled($container, $config['persistence']['orm'])
+            && !$this->isConfigEnabled($container, $config['persistence']['phpcr'])
+        );
+
         if (count($sonataBundles) && $config['sonata_admin_extension']['enabled']) {
             $this->loadSonataAdmin($config['sonata_admin_extension'], $loader, $container, $sonataBundles);
         }
--- a/Form/Type/SeoMetadataType.php
+++ b/Form/Type/SeoMetadataType.php
@@ -29,16 +29,23 @@ use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 class SeoMetadataType extends AbstractType
 {
     /**
-     * @var
+     * @var string
      */
     private $dataClass;
 
     /**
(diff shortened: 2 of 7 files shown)


WouterJ:
:+1: (maybe add a test to test if everything works correctly for both PHPCR + ORM?)

ElectricMaxxx:
There is a broken WebTest. Looks like an other Field is missing now.

Form/Type/SeoMetadataType.php
```diff
@@ -91,6 +91,7 @@ public function configureOptions(OptionsResolver $resolver)
             'data_class' => $this->dataClass,
             'translation_domain' => 'CmfSeoBundle',
             'required' => false,
+            'by_reference' => false,
```
dbu: adding this makes the test fail. i checked in the sandbox, and with PHPCR this leads to "A detached document was found through a child relationship during cascading a persist operation:  Symfony\\Cmf\\Bundle\\SeoBundle\\Doctrine\\Phpcr\\SeoMetadata@0000000073c1c05d0000000089dcaddb"

i will make the form type aware of whether we use ORM or not.

DependencyInjection/CmfSeoExtension.php
```diff
@@ -78,6 +78,11 @@ public function load(array $configs, ContainerBuilder $container)
             $sonataBundles[] = 'SonataDoctrineORMBundle';
         }
 
+        $container->setParameter($this->getAlias().'.form_mode_orm',
```
dbu: we can't use the backend_type_* parameters because they must only be set if true (the doctrine compiler pass only checks if the parameter exists, not the value of the parameter). and anyways, we want this logic in case somebody enabled both backends.

dbu:
okay, getting there. pushed some fixes for the --dev build.

dbu:
and its green! review anybody?

ElectricMaxxx:
Looks good to me. Nice hack with the orm parameter.

ElectricMaxxx:
@dbu  i would like to give you the honor of tagging and releasing :-) (ill at home with ill children next to me)

PR #262, opened 2015-09-06
Add locale meta tag with simple locale information

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #253 
| License       | MIT

Would just need to convert the locale information into RFC1766 standart language codes. But have no idiea how to do that, atm. The idea was to use something from `Intl::getLocaleBundle()`.

PR #278, opened 2016-04-02
Set Content-Language header instead of language meta tag

WouterJ:
The `%locale%` parameter is just a convention used in the Standard Edition. This bundle can not assume that this parameter is available. Instead, a custom config option should be introduced.

This bug was introduced in https://github.com/symfony-cmf/SeoBundle/pull/262. I found this while trying to update the CMF website.

/cc @dbu please merge this before releasing stable

### DIFF
--- a/DependencyInjection/CmfSeoExtension.php
+++ b/DependencyInjection/CmfSeoExtension.php
@@ -59,9 +59,9 @@ class CmfSeoExtension extends Extension
 
         $sonataBundles = array();
         if ($this->isConfigEnabled($container, $config['persistence']['phpcr'])) {
-            $container->setParameter($this->getAlias().'.backend_type_phpcr', true);
+            $container->setParameter('cmf_seo.backend_type_phpcr', true);
             $container->setParameter(
-                $this->getAlias().'.persistence.phpcr.manager_name',
+                'cmf_seo.persistence.phpcr.manager_name',
                 $config['persistence']['phpcr']['manager_name']
             );
             $sonataBundles[] = 'SonataDoctrinePHPCRAdminBundle';
@@ -70,15 +70,15 @@ class CmfSeoExtension extends Extension
         }
 
         if ($this->isConfigEnabled($container, $config['persistence']['orm'])) {
-            $container->setParameter($this->getAlias().'.backend_type_orm', true);
+            $container->setParameter('cmf_seo.backend_type_orm', true);
             $container->setParameter(
-                $this->getAlias().'.persistence.orm.manager_name',
+                'cmf_seo.persistence.orm.manager_name',
                 $config['persistence']['orm']['manager_name']
             );
             $sonataBundles[] = 'SonataDoctrineORMBundle';
         }
 
-        $container->setParameter($this->getAlias().'.form_mode_orm',
+        $container->setParameter('cmf_seo.form_mode_orm',
             $this->isConfigEnabled($container, $config['persistence']['orm'])
             && !$this->isConfigEnabled($container, $config['persistence']['phpcr'])
         );
@@ -139,7 +139,7 @@ class CmfSeoExtension extends Extension
             }
         }
 
-        $container->setParameter($this->getAlias().'.sonata_admin_extension.form_group', $config['form_group']);
+        $container->setParameter('cmf_seo.sonata_admin_extension.form_group', $config['form_group']);
         $loader->load('admin.xml');
     }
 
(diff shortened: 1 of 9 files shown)


WouterJ:
Oh, and I made the extension class to not use the `getAlias` method, but just `cmf_seo`. There is 0 change that we're going to change the alias and if we do, we need to do lots of find & replace stuff already.

DependencyInjection/Configuration.php
```diff
@@ -79,6 +79,7 @@ public function getConfigTreeBuilder()
                 ->scalarNode('title')->end()
                 ->scalarNode('description')->end()
                 ->scalarNode('original_route_pattern')->defaultValue(SeoPresentation::ORIGINAL_URL_CANONICAL)->end()
+                ->scalarNode('locale')->defaultNull()->end()
```
ElectricMaxxx: I do not see how to get the current locale now? Or what do we configure here? The key for the locale parameter in the routing or one of the locales?

DependencyInjection/CmfSeoExtension.php
```diff
@@ -59,9 +59,9 @@ public function load(array $configs, ContainerBuilder $container)
 
         $sonataBundles = array();
         if ($this->isConfigEnabled($container, $config['persistence']['phpcr'])) {
-            $container->setParameter($this->getAlias().'.backend_type_phpcr', true);
+            $container->setParameter('cmf_seo.backend_type_phpcr', true);
```
ElectricMaxxx: Why was that change needed? Thought that is the common way to use the bundle's alias.

DependencyInjection/Configuration.php
WouterJ: Either `'en'` or `%locale%` or anything else you want.

However, your question reveals something even more important: `%locale%` doesn't hold the current locale. It just holds the default locale. If we want to *current* locale, we need to use `$request->getLocale()`.

ElectricMaxxx: Oh then i will have to have a deeper look into the code where the Parameter is used.

WouterJ: The parameter is used to render the `http-equiv` language meta tag (see referenced PR).

ElectricMaxxx: Ohh then it would be better to really have the current locale ( in two capital letters as the Country Code).  
Now i remember that change - thought it is a cheap win to do that by the way. 
You can decide what to do with that: completely removing and opening an issue or fixing it by using the request object (be aware of the tests).

WouterJ:
After reading "[HTTP headers, meta elements and language information][1]" from W3C, I found out that the `language` meta tag doesn't exists. Instead, it should be `content-language`. Evenmore, this tag shouldn't be use either, as you should set the `Content-Language` header directly:

 > ### Quick Answer
 >
 >_**The HTTP `Content-Language` header can be used to provide metadata about the intended audience of the page, and can indicate that this is more than one language. The `Content-Language` value for an `http-equiv` attribute on a `meta` element should no longer be used.**_ You should use a language attribute on the html tag to declare the default language of the actual text in the page.

So I decided to remove this from the SeoPresentation and create a dedicated listener to set the correct header (I was surprised that Symfony didn't do this already!).

 [1]: https://www.w3.org/International/questions/qa-http-and-lang

DependencyInjection/CmfSeoExtension.php
WouterJ: See my first PR comment:

 > Oh, and I made the extension class to not use the `getAlias` method, but just `cmf_seo`. There is 0 change that we're going to change the alias and if we do, we need to do lots of find & replace stuff already.

ElectricMaxxx:
:+1:

EventListener/LanguageListener.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function onKernelResponse(FilterResponseEvent $event)
+    {
+        if ($event->getResponse()->headers->has('Content-Language')) {
+            return;
+        }
+
+        $locale = $event->getRequest()->getLocale();
+        $language = current(explode('_', $locale, 2));
+        $event->getResponse()->headers->set('Content-Language', $language);
```
ElectricMaxxx: But that work isn't wrong, right? I would keep that as you have invested the time.

ElectricMaxxx:
Great work @WouterJ  Can you fix the style issue, please. I will merge than.

DependencyInjection/CmfSeoExtension.php
ElectricMaxxx: if i would have understood that comment before, i wouldn't have asked. :-)

dbu:
fixed styling and added a changelog entry.

looks good to merge to me. @ElectricMaxxx do you want to hit the button and tag the release?

Issue #280, opened 2016-04-07
[Sitemap] add depth information to the UrlInformation


ElectricMaxxx:
As a visitor of a html sitemap i want to see the sitmap with indents of the position of the content's routes in the path to get a better overview.

NOTE:
We should do that by passing, the route node's depth property, we hopefully are able to acces.

AC:
- depth property is available in the templates
- \* html and json template are using that

dbu:
sounds like a good idea to me. maybe we should "normalize" the depth so
that the route root is at 0. that would mean take the depth of the root
and substract that from the depth we pass to the templating.

even the xml template could include that attribute, as long as that does
not violate some schema.

On 07.04.2016 08:52, Maximilian Berghoff wrote:
> As a visitor of a html sitemap i want to see the sitmap with indents of
> the position of the content's routes in the path to get a better overview.
> 
> NOTE:
> We should do that by passing, the route node's depth property, we
> hopefully are able to acces.
> 
> AC:
> 
>   * depth property is available in the templates
>   * * html and json template are using that
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/symfony-cmf/SeoBundle/issues/280>
> 

-- 
Liip AG // Agile Web Development // T +41 43 500 39 80
CH-8005 Zurich // PGP 0xA581808B // www.liip.ch

ElectricMaxxx:
The items in the `<urlsets></urlsets>` have no semantics. It is just a plain list for google, except somebody is using its own xml-template for custom usage. For that case i would add the information in general to the parameters passed to the template.
I am thinking what is faster: Either simply counting by imploding by `/` or grabing the node and try to get its depth information. 

Btw: is there a way to order by node path, when fetching the documents? So we will have the result a little bit aligned without sorting by the parents.

ElectricMaxxx:
Maybe we can put that one on a 1.2.1? I would like to implement the sitemap on our website (later sandbox too) and would need that feature.

dbu:
> The items in the |<urlsets></urlsets>| have no semantics. It is just a
> plain list for google, except somebody is using its own xml-template for
> custom usage. For that case i would add the information in general to
> the parameters passed to the template.

okay, fair enough.

> I am thinking what is faster: Either simply counting by imploding by |/|
> or grabing the node and try to get its depth information.

if we operate on the documents, i think we can ask the document manager
[...]

PR #282, opened 2016-04-12
implement and register DepthGuesser

ElectricMaxxx:
This PR will implement a solution to implement #280 I did that by adding a so called `DepthGuesser` for enabled phpcr. The depth is 0 for the first entry on content_basepath. So a content on  `/cms/content/test` with basepath `/cms/content` will get a 0 as result. The first ones will get no indend in the html template doing it that way.

btw: i miss the nice and helpfull PR template in the guidelines, which i have copied every time.

### DIFF
--- a/Doctrine/Phpcr/SitemapDocumentProvider.php
+++ b/Doctrine/Phpcr/SitemapDocumentProvider.php
@@ -42,7 +42,7 @@ class SitemapDocumentProvider implements LoaderInterface
     {
         // todo rewrite query when https://github.com/symfony-cmf/CoreBundle/issues/126 is ready
         $documentsCollection = $this->manager->createQuery(
-            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true)',
+            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true) ORDER By id',
             QueryInterface::JCR_SQL2
         )->execute();
 
--- a/Doctrine/Phpcr/SitemapDocumentProvider.php
+++ b/Doctrine/Phpcr/SitemapDocumentProvider.php
@@ -42,7 +42,7 @@ class SitemapDocumentProvider implements LoaderInterface
     {
         // todo rewrite query when https://github.com/symfony-cmf/CoreBundle/issues/126 is ready
         $documentsCollection = $this->manager->createQuery(
-            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true) ORDER By id',
+            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true) ORDER By [jcr:path]',
             QueryInterface::JCR_SQL2
         )->execute();
 
--- a/Model/UrlInformation.php
+++ b/Model/UrlInformation.php
@@ -60,6 +60,11 @@ class UrlInformation
      */
     private $alternateLocales;
 
+    /**
+     * @var int
+     */
+    private $depth;
+
     public function __construct()
     {
         $this->alternateLocales = array();
@@ -94,6 +99,7 @@ class UrlInformation
             'lastmod' => $this->lastModification,
             'priority' => $this->priority,
             'alternate_locales' => array(),
+            'depth' => $this->depth,
         );
         foreach ($result as $key => $value) {
             if (null === $value) {
@@ -247,5 +253,21 @@ class UrlInformation
         $this->alternateLocales[] = $alternateLocale;
 
         return $this;
+    }
+
+    /**
+     * @return int
+     */
+    public function getDepth()
+    {
+        return $this->depth;
+    }
+
+    /**
+     * @param int $depth
+     */
+    public function setDepth($depth)
+    {
+        $this->depth = $depth;
     }
 }
--- a/Model/UrlInformation.php
+++ b/Model/UrlInformation.php
@@ -61,7 +61,7 @@ class UrlInformation
     private $alternateLocales;
 
     /**
-     * @var int
+     * @var int|null
      */
     private $depth;
 
@@ -264,7 +264,7 @@ class UrlInformation
     }
 
     /**
-     * @param int $depth
+     * @param int|null $depth
      */
     public function setDepth($depth)
     {
--- a/Resources/config/phpcr-sitemap.xml
+++ b/Resources/config/phpcr-sitemap.xml
@@ -15,5 +15,11 @@
 
             <tag name="cmf_seo.sitemap.loader" priority="-2"/>
         </service>
+        <service id="cmf_seo.sitemap.phpcr.depth_guesser" class="Symfony\Cmf\Bundle\SeoBundle\Sitemap\DepthGuesser">
+            <argument type="service" id="doctrine_phpcr"/>
+            <argument>%cmf_routing.dynamic.persistence.phpcr.content_basepath%</argument>
+
+            <tag name="cmf_seo.sitemap.guesser" position="-2"/>
+        </service>
     </services>
 </container>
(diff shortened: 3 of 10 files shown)


ElectricMaxxx:
uhhh the style check is broken, cause i used actual licence information in the new files. :-(

Doctrine/Phpcr/SitemapDocumentProvider.php
```diff
@@ -42,7 +42,7 @@ public function load($sitemap)
     {
         // todo rewrite query when https://github.com/symfony-cmf/CoreBundle/issues/126 is ready
         $documentsCollection = $this->manager->createQuery(
-            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true)',
+            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true) ORDER By id',
```
dbu: is there a property `id`? usually there is not, and the node path is mapped onto a class field $id. but you can't sort by the path afaik. did you verify that this works? in the search result, the field is called [jcr:path] but i doubt that that would work either.

Model/UrlInformation.php
```diff
@@ -60,6 +60,11 @@ class UrlInformation
      */
     private $alternateLocales;
 
+    /**
+     * @var int
+     */
+    private $depth;
```
dbu: should we default this to 0 so its always an integer?

Resources/views/Sitemap/index.html.twig
```diff
@@ -1,7 +1,7 @@
 <ul class="cmf-sitemap">
     {% for url in urls %}
         <li>
-            <a href="{{ url.location }}" title="{{ url.label }}">{{ url.label }}</a>
+            {% if url.depth is not null %} <span class="indent-{{ url.depth }}"></span>{% endif %}<a href="{{ url.location }}" title="{{ url.label }}">{{ url.label }}</a>
```
dbu: i am no frontend expert but i would have expected the class to be on the `<li>` element rather than a span. if span, it should probably wrap the `<a>` rather than just be before it for cleaner styling options.

Sitemap/DepthGuesser.php
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * DepthGuesser constructor.
+     *
+     * @param ManagerRegistry $managerRegistry
+     */
+    public function __construct(ManagerRegistry $managerRegistry, $contentBasePath)
+    {
+        $this->managerRegistry = $managerRegistry;
+        $exploded = explode('/', $contentBasePath);
+        $this->offset = count($exploded);
```
dbu: why not substr_count?

Model/UrlInformation.php
dbu: okay, null means undefined, makes sense below

Doctrine/Phpcr/SitemapDocumentProvider.php
ElectricMaxxx: The reason why i changed that is to have all children of the the same parent alligned in the same indent block.

Model/UrlInformation.php
```diff
@@ -248,4 +254,20 @@ public function addAlternateLocale(AlternateLocale $alternateLocale)
 
         return $this;
     }
+
+    /**
+     * @return int
```
dbu: lets make this `@return int|null The 0-based depth value relative to the base path. Or null if depth is not known.

Model/UrlInformation.php
ElectricMaxxx: i have som `null === $depth` checks, but you are right i think.

Resources/views/Sitemap/index.html.twig
ElectricMaxxx: i am not that frontend expert too, but i will ask one at the next door :-)

composer.json
```diff
@@ -44,7 +44,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev"
+            "dev-master": "1.3-dev"
```
dbu: are you sure about this one? did we not start some BC breaks in master already?

Sitemap/DepthGuesser.php
ElectricMaxxx: uhh nice function, never heard of that :-)

dbu:
> uhhh the style check is broken, cause i used actual licence information
> in the new files. :-(

i changed the file header on styleci to 2016 and let it fix the master
branch. you now need to rebase this branch on master or all other files
will fail the stylecheck...

regarding the pull request template: github recently announced they
support that by adding a magic file to the repository root. i think it
would be a good idea to add that everywhere. maybe check if
symfony/symfony has one and take that one - afaik we use pretty much the
same template.

Doctrine/Phpcr/SitemapDocumentProvider.php
dbu: it does make sense to order the pages by path, i am just not sure if
PHPCR can do it, or if it can if this is the way to achieve it.

composer.json
ElectricMaxxx: That is the way we discussed it in the issue

Model/UrlInformation.php
ElectricMaxxx: I will do it with `0` at all places. So we have one type only :-)

composer.json
dbu: i propose we start a 1.3 branch from the 1.2.0 tag. that way we don't revert what is in master already.

Resources/views/Sitemap/index.html.twig
ElectricMaxxx: The experts said: 

To semanticaly put it in the right HTML structure we should create one `<ul></ul>` for each block of children with the same parent (and the same depth). The easy way is to put the `indent-X` classes on the `<li>`element.

Model/UrlInformation.php
dbu: i think null actually makes sense. otherwise you don't know if depth was
set to 0 or not set.

Resources/views/Sitemap/index.html.twig
dbu: lets put the class on the <li>. the <ul> would be really difficult as we
would need to look-ahead somehow.

Resources/config/phpcr-sitemap.xml
```diff
@@ -15,5 +15,11 @@
 
             <tag name="cmf_seo.sitemap.loader" priority="-2"/>
         </service>
+        <service id="cmf_seo.sitemap.phpcr.depth_guesser" class="Symfony\Cmf\Bundle\SeoBundle\Sitemap\DepthGuesser">
+            <argument type="service" id="doctrine_phpcr"/>
+            <argument>%cmf_routing.dynamic.persistence.phpcr.content_basepath%</argument>
+
+            <tag name="cmf_seo.sitemap.guesser" position="-2"/>
```
WouterJ: -2 is very close to the default. Let's use -20 or so.

ElectricMaxxx: We have got the same for all other default ones.

Resources/views/Sitemap/index.html.twig
ElectricMaxxx: sure, we would have sorted the urlInformation into those blocks.

Resources/config/phpcr-sitemap.xml
dbu: but the other guessers are at -2 as well. which maybe is not a great
choice but changing that would be a BC break. i think its less confusing
to keep this in line with the others.

people that need to extend can always go to positive or higher negative
numbers. from the cmf side, nothing is at the default priority.

ElectricMaxxx:
Closed by #284

PR #284, opened 2016-04-13
Depth information in sitemap into 1.2

ElectricMaxxx:
#282 done on branch 1.2 to create a new 1.3 version out of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+1.3.0 (unreleased)
+------------------
+
+* **2016-04-12**: Build and register DepthGuesser for PHPCR to use depths information for structure sitemap 
+
 1.2.0
 -----
 
--- a/Doctrine/Phpcr/SitemapDocumentProvider.php
+++ b/Doctrine/Phpcr/SitemapDocumentProvider.php
@@ -42,7 +42,7 @@ class SitemapDocumentProvider implements LoaderInterface
     {
         // todo rewrite query when https://github.com/symfony-cmf/CoreBundle/issues/126 is ready
         $documentsCollection = $this->manager->createQuery(
-            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true)',
+            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true) ORDER BY [jcr:path]',
             QueryInterface::JCR_SQL2
         )->execute();
 
--- a/Sitemap/DepthGuesser.php
+++ b/Sitemap/DepthGuesser.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2016 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Sitemap;
+
+use Doctrine\Bundle\PHPCRBundle\ManagerRegistry;
+use Doctrine\Common\Util\ClassUtils;
+use Doctrine\ODM\PHPCR\DocumentManager;
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+
+/**
+ * This guesser will add the depth of a document persisted on a phpcr node.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class DepthGuesser implements GuesserInterface
+{
+    /**
+     * @var ManagerRegistry
+     */
+    private $managerRegistry;
+
+    /**
+     * @var int The depth of the content base path as the depth offset.
+     */
+    private $offset;
+
+    /**
+     * DepthGuesser constructor.
+     *
+     * @param ManagerRegistry $managerRegistry
+     * @param $contentBasePath
+     */
+    public function __construct(ManagerRegistry $managerRegistry, $contentBasePath)
+    {
+        $this->managerRegistry = $managerRegistry;
+        $this->offset = ('/' === $contentBasePath) ? 1 : substr_count($contentBasePath, '/') + 1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function guessValues(UrlInformation $urlInformation, $object, $sitemap)
+    {
+        if (null !== $urlInformation->getDepth()) {
+            return;
+        }
+
+        $manager = $this->managerRegistry->getManagerForClass(ClassUtils::getRealClass(get_class($object)));
+        if (!$manager instanceof DocumentManager) {
+            return;
+        }
+
+        $node = $manager->getNodeForDocument($object);
+        if (null === $node) {
+            return;
+        }
+        $urlInformation->setDepth($node->getDepth() - $this->offset);
+    }
+}
--- a/Tests/Functional/Doctrine/Phpcr/SitemapDocumentProviderTest.php
+++ b/Tests/Functional/Doctrine/Phpcr/SitemapDocumentProviderTest.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2016 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 namespace Symfony\Cmf\Bundle\SeoBundle\Tests\Functional\Doctrine\Phpcr;
 
 use Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr\SitemapDocumentProvider;
--- a/Tests/Functional/Doctrine/Phpcr/SitemapDocumentProviderTest.php
+++ b/Tests/Functional/Doctrine/Phpcr/SitemapDocumentProviderTest.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Tests\Functional\Doctrine\Phpcr;
+
+use Symfony\Cmf\Bundle\SeoBundle\Doctrine\Phpcr\SitemapDocumentProvider;
+use Symfony\Cmf\Component\Testing\Functional\BaseTestCase;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class SitemapDocumentProviderTest extends BaseTestCase
+{
+    /**
+     * @var SitemapDocumentProvider
+     */
+    private $documentProvider;
+
+    public function setUp()
+    {
+        $this->db('PHPCR')->createTestNode();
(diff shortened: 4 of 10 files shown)


Doctrine/Phpcr/SitemapDocumentProvider.php
```diff
@@ -42,7 +42,7 @@ public function load($sitemap)
     {
         // todo rewrite query when https://github.com/symfony-cmf/CoreBundle/issues/126 is ready
         $documentsCollection = $this->manager->createQuery(
-            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true)',
+            'SELECT * FROM [nt:unstructured] WHERE (visible_for_sitemap = true) ORDER By [jcr:path]',
```
ElectricMaxxx: @dbu created a special function test for that. That test is a green test, but i inserted `DESC` and got the fail.

Sitemap/DepthGuesser.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * DepthGuesser constructor.
+     *
+     * @param ManagerRegistry $managerRegistry
+     */
+    public function __construct(ManagerRegistry $managerRegistry, $contentBasePath)
+    {
+        $this->managerRegistry = $managerRegistry;
+        $this->offset = substr_count($contentBasePath, '/') + 1;
```
ElectricMaxxx: `+1` to get a `0` on root element.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Changelog
 =========
```
dbu: can you please add `1.3.0 (unreleased)` as title? i fear we might get confused between 2.0 and 1.3 otherwise.

Doctrine/Phpcr/SitemapDocumentProvider.php
dbu: if you omit the order by, does the test succeed as well?

Sitemap/DepthGuesser.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function guessValues(UrlInformation $urlInformation, $object, $sitemap)
+    {
+        if (null !== $urlInformation->getDepth()) {
+            return;
+        }
+
+        $manager = $this->managerRegistry->getManagerForClass(get_class($object));
```
dbu: i think we should use ClassUtils::getRealClass instead of get_class here, in case $object is a proxy of a document.

Doctrine/Phpcr/SitemapDocumentProvider.php
ElectricMaxxx: You mean if that is order is the default? I will look at it as well.

Sitemap/DepthGuesser.php
dbu: i am not sure if + 1 is correct.

/ => 0
/cms => 1
/cms/content => 2

or am i miscounting here? i would expect the PHPCR depth to count from 0 as well. if base is /cms/content and i look at the depth of /cms/content, it would be 2, minus the 2 slashes i end up with 0.

we should handle the special case `'/' === $contentBasePath` btw.

Doctrine/Phpcr/SitemapDocumentProvider.php
dbu: in general i expect there to be no guaranteed order without an order by
clause. the last section on this doc is rather confusing - but i think
its about a product not using the normal jackrabbit, so no idea how
relevant it is:

https://www.exoplatform.com/docs/public/index.jsp?topic=%2FPLF35%2FJCR.SearchingRepositoryContent.OrderingSpecifying.html

how did ORDER BY [jcr:path] DESC fail? invalid query or just different
html? the order of the output would be reverted if it works...

ElectricMaxxx: The `ORDER BY [jcr:path] DESC` did not fail. The test on the sorting order of the documents (some changed files after this one) fails.

Tests/Functional/Doctrine/Phpcr/SitemapDocumentProviderTest.php
```diff
@@ -0,0 +1,48 @@
...
+    public function testDocumentOrder()
+    {
+        $documents = $this->documentProvider->load('default');
+
+        $paths = array();
+        foreach ($documents as $document) {
+            $paths[] = $document->getId();
+        }
+
+        $this->assertEquals(
```
ElectricMaxxx: @dbu  this a a hard coded test on the sorting order of the document's path i get by the provider and its query.

Sitemap/DepthGuesser.php
ElectricMaxxx: :+1:

Sitemap/DepthGuesser.php
ElectricMaxxx: contentBasePath: /cms/content => offset = 2 +1, contentDocumentPath: /cms/content/test-document => depth = 3 ==> 3 - 3 = 0 => the root on the contentpath.

contentBasePath: /cms/content => offset = 2 +1, contentDocumentPath: /cms/content/collection/test-document => depth = 4 ==> 4 - 3 = 1 => the content will be dispplayed with an indent of 1

contentBasePath: / => offset = 1 +1 = 2, contentDocumentPath: /test-document => depth = 1 ==> 1 - 2 = 0 => @dbu you are right :-)

CHANGELOG.md
ElectricMaxxx: normally we add the title when releasing

dbu: i have seen the thing with "x.y.0 (unreleased)" in other projects and
found it quite useful. it will come in handy when we merge 1.3 to master
so that we don't confuse this section with the 2.0 changes section.

Doctrine/Phpcr/SitemapDocumentProvider.php
dbu: ah ok, well DESC would reverse the order, so its normal that the order is not the same :-)

lets keep this then. can you please fix "s/By/BY" just to satisfy my pedantism?

Sitemap/DepthGuesser.php
dbu: for / i am right that we need special handling, but for the rest you are right. i thought /cms/content itself could be a document that we handle, but this is not the case as its created as a generic node to have somewhere to attach documents. so `$this->offset = ('/' === $contentBasePath) ? 1 : substr_count($contentBasePath, '/') + 1;

CHANGELOG.md
ElectricMaxxx: I am totally on your site.

WouterJ:
I don't really get why this is on 1.2. Are we going to add a new feature in 1.2.1? Is this feature of such importance or is it a bug fixing feature?

If we want to release it as 1.3, we have to create a 1.3 branch from the 1.2 branch *now* and then target this PR to the 1.3 branch. Otherwise, this feature ends up in the 1.2 branch.

ElectricMaxxx:
@WouterJ Reason: @dbu mentioned on the ticket, that the change of the `index.html.twig` is a BC break. When you don't agree i can live with 1.2.1.

Sitemap/DepthGuesser.php
ElectricMaxxx: OK, did it and wrote test for that edge case too.

CHANGELOG.md
ElectricMaxxx: Your comment it one line to high.

dbu:
i don't think we have a BC break here. but its a new feature. so strictly speaking it should be 1.3 (or 2.0, but we said doing a 1.3 soonish with non-breaking changes would be ok).

indeed wouter is right, the branch should be called 1.3, as 1.2 is for maintenance of the 1.2 version.

ElectricMaxxx:
@dbu i do not understand what i have to change on the PR after your discussion.

btw: travis has a licence header style problem on this branch too :-)

WouterJ:
> but we said doing a 1.3 soonish with non-breaking changes would be ok

Isn't it a bit strange to release a new minor with just one new feature?

dbu:
>     but we said doing a 1.3 soonish with non-breaking changes would be ok
> 
> Isn't it a bit strange to release a new minor with just one new feature?

i don't see a problem releasing it.

i would propose to squash the commits, rebase on 1.2 and then create the
1.3 branch from 1.2 and open a new PR of this branch to 1.3. but if
wouter is unhappy with that, we can either smuggle it into 1.2.1 or
merge it into master and wait until 2.0 with releasing.

regarding styleci: i only did the fix on master. now done it on the 1.2
branch too and pushed. you will need to rebase this branch on 1.2 to get
them in.

ElectricMaxxx:
i rebased on 1.2 and squashed all commits.

dbu:
styleci looks happy now \o/

@WouterJ how do you see this?

dbu:
styleci looks happy now \o/

@WouterJ how do you see this question of which release this can go into?

WouterJ:
Definitely not 1.2.1 imo, for the rest I don't care much. (please note that this PR is targetting 1.2 so it should not be merged)

ElectricMaxxx:
@WouterJ it is should be merged into 1.2 and 1.3 into master later, cause maste have got changes after release.

dbu:
if we want the new feature to correctly only go to a 1.3 branch, it needs to be merged into a 1.3 branch that we create from the 1.2 branch. and then we can merge 1.3 into master once this is merged to 1.3.

ElectricMaxxx:
To get the stuff a clear in my head (and in the one of git too *g*):

* i create a 1.3 branch out of 1.2
* i create a PR into 1.3
* you will merge that one
* i tag the new release (or a RC before?)
* i merge 1.3 into master and will get merge conflicts as the licence change is in both branches

right?

dbu:
mostly.
i would not release 1.3 right away, but wait a few weeks if something else comes up. we could also use the 1.3 branch to start deprecating things as we refactor in master. and target mid may for the 1.3 release. wdyt?

ElectricMaxxx:
sounds like a plan

ElectricMaxxx:
@rdohms you told me yesterday, that i can make bug fixes and really small stuff on the patch part of a semver version. When is a feature/change to big that i need a new minor version?
We have an open diskussion on my PR in here. The feature is just an information, that i missed while implementing the sitemap feaature for version 1.2 - the depth. This is created and passed to the template (default one, mostly overwritten) now. there the change is minimal to, just one class entry, which does nothing when there is no css styling for that.
I don't wanna wait for that change until our next release (major this time) is there, cause i have got an open [PR](https://github.com/symfony-cmf/symfony-cmf-website/pull/94) on our website repository..

ElectricMaxxx:
No answer? so i will do as proposed by @dbu

dbu:
i would still be ok to put it into 1.2, as its strictly an addition, and we don't touch code that can be replaced / overwritten. the additional css class has no effect until styling rules are added. 

but otherwise, i suggest you just create the 1.3 branch from this branch, push that and close the PR. then we have the 1.3 version.

ElectricMaxxx:
closed due to #286

PR #286, opened 2016-04-22
Depth information in sitemap into 1.3

ElectricMaxxx:
followup from #284


dbu:
here something went wrong. the 2.0 commits slipped into this branch, plus you try to do a PR against master which already represents 2.0. i would simply create a branch from the depth information branch (without the commits from master that sliiped to you now) and call that branch 1.3.

ElectricMaxxx:
cause i am stupid :-(

PR #289, opened 2016-04-22
add github template from symfony

ElectricMaxxx:
I just picked the symfony template as they are very close to us. Started here and will add them everywhere when we are fine with that. Should we add a notice to the contribution guidelines?

Information: https://github.com/blog/2111-issue-and-pull-request-templates

### DIFF
--- a/.github/PULL_REQUEST_TEMPLAE.md
+++ b/.github/PULL_REQUEST_TEMPLAE.md
@@ -1,0 +1,11 @@
+| Q             | A
+| ------------- | ---
+| Branch?       | "master" for new features / 1.0, 1.1, 1.2, 1.3, 1.4 for fixes
+| Bug fix?      | yes/no
+| New feature?  | yes/no
+| BC breaks?    | yes/no
+| Deprecations? | yes/no
+| Tests pass?   | yes/no
+| Fixed tickets | comma-separated list of tickets fixed by the PR, if any
+| License       | MIT
+| Doc PR        | reference to the documentation PR, if any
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLAE.md
```diff
@@ -0,0 +1,11 @@
+| Q             | A
+| ------------- | ---
+| Branch?       | "master" for new features / 1.0, 1.1, 1.2, 1.3, 1.4 for fixes
```
dbu: i would want to avoid listing all branches here. we would need to maintain that constantly. rather say "the branch of the current release". (also, we don't maintain 1.0 anymore, bugfixes would go into 1.2 atm)

dbu:
@WouterJ do you agree with this? did we have a template of our own somewhere?

.github/PULL_REQUEST_TEMPLAE.md
```diff
@@ -0,0 +1,11 @@
...
+| Bug fix?      | yes/no
+| New feature?  | yes/no
+| BC breaks?    | yes/no
+| Deprecations? | yes/no
+| Tests pass?   | yes/no
```
WouterJ: This is simply replaced by Travis build, right?

.github/PULL_REQUEST_TEMPLAE.md
```diff
@@ -0,0 +1,11 @@
...
+| Fixed tickets | comma-separated list of tickets fixed by the PR, if any
+| License       | MIT
```
WouterJ: Do we really need to include this?

dbu: IANAL but i think it might help legally.

.github/PULL_REQUEST_TEMPLAE.md
dbu: indeed, i always find that weird with symfony PRs. we could simply remove this line imho.

ElectricMaxxx:
btw: https://github.com/symfony-cmf/routing-bundle/blob/master/.github/PULL_REQUEST_TEMPLATE.md

ElectricMaxxx:
So you like it now?

PR #290, opened 2016-04-22
fix build status icon

ElectricMaxxx:
travis changed the url
(Same for all bundles)


dbu:
ah of course, because we renamed the repos... could you do that for all the cmf bundles, maybe? push directly to master would be fine if you only change that bit.

ElectricMaxxx:
:+1: tonight.

ElectricMaxxx:
@dbu not at all they moved from `http://secure.` to `https://`.

dbu:
ah ok. thanks for the fixing round!

Issue #292, opened 2016-05-25
[Sitemap] Provide default last modification date guesser


WouterJ:
I think this should be introduced by calling a custom method or the like (maybe PHPCR can give us the last mod date already?).

ElectricMaxxx:
@dantleech can you give us some feedback if a node has such information,
and maybe how to get it through phpcr-odm?

dbu:
http://doctrine-orm.readthedocs.io/projects/doctrine-phpcr-odm/en/latest/cookbook/last-modified.html

ElectricMaxxx:
You whant it on the 1.3, that we still have to release. Or wait for 2.0?

dbu:
if somebody has time to work on it, it would be nice in 1.3. its no BC break, just an additional guesser.

ElectricMaxxx:
I think i can do it after my REST Talk next week. Now the waiting for the release makes Sense, that is a sweet new feature.

ElectricMaxxx:
But one question: That information could also go into the header for
caching or the FosHttCacheBundle could use that information, right?

dbu:
in FOSHttpCache, we have no doctrine integrations. we don't have a guesser architecture there. the problem doing that would be to know which object to use. see also https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/184 for a lot of discussion about content object integration.

PR #295, opened 2016-06-07
implement last modification date guesser

ElectricMaxxx:
Solves #292

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changelog
 1.3.0 (unreleased)
 ------------------
 
+* **2016-06-09**: Build and register LastModificationGuesser for PHPCR to use a last modification date on a sitemap. 
 * **2016-04-12**: Build and register DepthGuesser for PHPCR to use depths information for structure sitemap 
 
 1.2.0
--- a/Resources/config/phpcr-sitemap.xml
+++ b/Resources/config/phpcr-sitemap.xml
@@ -21,5 +21,13 @@
 
             <tag name="cmf_seo.sitemap.guesser" position="-2"/>
         </service>
+
+        <service
+                id="cmf_seo.sitemap.phpcr.last_modified_guesser"
+                class="Symfony\Cmf\Bundle\SeoBundle\Sitemap\LastModifiedGuesser">
+            <argument type="service" id="doctrine_phpcr" />
+
+            <tag name="cmf_seo.sitemap.guesser" position="-2"/>
+        </service>
     </services>
 </container>
--- a/Sitemap/LastModifiedGuesser.php
+++ b/Sitemap/LastModifiedGuesser.php
@@ -1,0 +1,89 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2016 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\SeoBundle\Sitemap;
+
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\Common\Util\ClassUtils;
+use Doctrine\ODM\PHPCR\DocumentManager;
+use Doctrine\ODM\PHPCR\Mapping\ClassMetadata;
+use Symfony\Cmf\Bundle\SeoBundle\Model\UrlInformation;
+
+/**
+ * This guesser will add last modified date of an document to the url information, that can be rendered
+ * to the sitemap.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class LastModifiedGuesser implements GuesserInterface
+{
+    /**
+     * @var ManagerRegistry
+     */
+    private $managerRegistry;
+
+    /**
+     * LastModifiedGuesser constructor.
+     *
+     * @param ManagerRegistry $manager
+     */
+    public function __construct(ManagerRegistry $manager)
+    {
+        $this->managerRegistry = $manager;
+    }
+
+    /**
+     * Updates UrlInformation with new values if they are not already set.
+     *
+     * @param UrlInformation $urlInformation The value object to update.
+     * @param object         $object         The sitemap element to get values from.
+     * @param string         $sitemap        Name of the sitemap being built.
+     */
+    public function guessValues(UrlInformation $urlInformation, $object, $sitemap)
+    {
+        if (null !== $urlInformation->getLastModification()) {
+            return;
+        }
+
+        $className = ClassUtils::getRealClass(get_class($object));
+        $manager = $this->managerRegistry->getManagerForClass($className);
+        if (!$manager instanceof DocumentManager) {
+            return;
+        }
+
+        /** @var ClassMetadata $metadata */
+        $metadata = $manager->getClassMetadata($className);
+        $mixins = $metadata->getMixins();
+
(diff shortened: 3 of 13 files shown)


Sitemap/LastModifiedGuesser.php
```diff
@@ -0,0 +1,84 @@
...
+     * @return null
+     */
+    public function guessValues(UrlInformation $urlInformation, $object, $sitemap)
+    {
+        if (null !== $urlInformation->getLastModification()) {
+            return null;
+        }
+        
+        $className = ClassUtils::getRealClass(get_class($object));
+        $manager = $this->managerRegistry->getManagerForClass($className);
```
dbu: what does this do for a completely unmanaged class?

Sitemap/LastModifiedGuesser.php
```diff
@@ -0,0 +1,84 @@
...
+        $mixins = $metadata->getMixins();
+
+        if (!in_array('mix:lastModified', $mixins)) {
+            return null;
+        }
+
+        $fields = array_filter($metadata->getFieldNames(), function ($fieldName) use ($metadata) {
+            $field = $metadata->getField($fieldName);
+
+            return 'jcr:lastModified' == $field['property'];
```
dbu: i would move this into a private function instead of an array filter. then you can return the field as soon as you find it, instead of looping through the rest of the array.

Tests/Functional/Doctrine/Phpcr/SitemapDocumentProviderTest.php
```diff
@@ -48,6 +48,7 @@ public function testDocumentOrder()
         $this->assertEquals(
             array(
                 '/test/content/sitemap-aware',
+                1 => '/test/content/sitemap-aware-last-mod-date',
```
dbu: why have an explicit key here?

Resources/config/phpcr-sitemap.xml
```diff
@@ -21,5 +21,13 @@
 
             <tag name="cmf_seo.sitemap.guesser" position="-2"/>
         </service>
+
+        <service
+                id="cmf_seo.sitemap.phpcr.last_mod_date_guesser"
```
dbu: i would call this ... .last_modified_guesser to conform with the class name.

Tests/Functional/Doctrine/Phpcr/SitemapDocumentProviderTest.php
ElectricMaxxx: Stupid copy from Test expectation. Whant to fix it and my power goes down just before boarding. Will do the fixes later.

Tests/Unit/Sitemap/LastModifiedGuesserTest.php
```diff
@@ -0,0 +1,115 @@
...
+        return array('LastModification');
+    }
+
+    public function testGuessNoOverwrite()
+    {
+        $urlInformation = new UrlInformation();
+        $urlInformation->setLastModification(new \DateTime('2016-06-06'));
+
+        $this->guesser->guessValues($urlInformation, $this->data, 'default');
+        $this->assertEquals('2016-06-06T00:00:00+02:00', $urlInformation->getLastModification());
```
ElectricMaxxx: I think i should expect those tests to break :-(

dbu: set the timezone in new \DateTime?

ElectricMaxxx: @dbu it is not only the timezone, i have got a difference of 2 seconds
on travis. So i have to set the complete time.

ElectricMaxxx: ahh ..
You are right. The 2 is on the timezone :-) sorry

Tests/WebTest/SitemapTest.php
```diff
@@ -87,14 +90,19 @@ public function getFormats()
                          <xhtml:link href="http://localhost/sitemap-aware?_locale=de" hreflang="de" rel="alternate"/>
                      </url>
                      <url>
+                         <loc>http://localhost/sitemap-aware-last-mod-date</loc>
+                         <lastmod>2016-07-06T00:00:00+02:00</lastmod>
```
ElectricMaxxx: That value is created in https://github.com/symfony-cmf/seo-bundle/pull/295/files#diff-c2db5dda6526964e483d707ab1118506R86 Same way we do it for the unit-tests. But the travis tests says `2016-07-05T22:00:00+00:00` same time but other timezone. @dbu you know what happens on phpcr-odm when saving DateTime?

ElectricMaxxx: @heiglandreas is that something for you?

dbu: i think you should not try to compare rendered dates, but the datetime objects themselves. not sure if a datetime object stores the timezone and if `==` looks at the timezone. otherwise get the timestamp from the two datetime objects.

ElectricMaxxx: If will be very hard here, cause match the response string (it is a webtest). Should we Switch to a preg_match of strings instead of the complete strings.

dbu: maybe you can explicitly set the timezone to Europe/Berlin in your
template? or in the autoload/bootstrap file?

surprised that this did not turn up on travis-ci. but maybe they are
dutch and thus in the same timezone as us...

heiglandreas: Looks like Travis runs with default_timezone set to UTC. Where is the ```lastmod``` parameter set? I might just be to old to find it ;)

ElectricMaxxx: There is a link some comments above in the `LoadContent`.

heiglandreas: I'm not *that* old! I was more refering to where the actual data is set as I couldn'T figure out at first where that happens. The referenced line is the point where you set the comparison date but the question is more where the value is set that you are comparing against. Whatever: I think I found it and it get's fed from the database. But where is that database-value set in the first place? As that's where the offset-information is stored.

As the question is **when**  the value was changed and you can't reliably check **where** (geographically) it was changed, I'd always store the value in UTC and execute all the tests against it using ```new DateTimezone('UTC')``` for checks. I'D also check whether the dates are always written as UTC to the database.

ElectricMaxxx: looks good now. manually set the timezone now.

dbu: that makes sense. afaik phpcr is supposed to retain the timezone. and when you run those tests in a different timezone, thinks become shaky.

dbu:
coming good. @ElectricMaxxx when you think its ready, i can have a last look over the code and will squash-merge it.

ElectricMaxxx:
@dbu I think it is ready. i am ready for you review.

dbu:
looks good. one last bit: you should add a note in the CHANGELOG.md file.

ElectricMaxxx:
@dbu done.

ElectricMaxxx:
i will squash when you like the text.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Changelog
 1.3.0 (unreleased)
 ------------------
 
+* **2016-04-12**: Build and register LastModificationGuesser for PHPCR to use a last modification date on a sitemap.
```
dbu: i like the text but the date is wrong :-)

ElectricMaxxx:
fixed, squashed and ready to merge now

dbu:
:confetti_ball:

PR #303, opened 2016-07-04
Use route_basepaths parameters instead of deprecated route_basepath one

WouterJ:


Doctrine/Phpcr/BaseSuggestionProvider.php
```diff
@@ -56,4 +57,26 @@ public function getManagerForClass($class)
...
+     * @return null|\Traversable
+     */
+    protected function findParentRoute($requestedPath)
+    {
+        $manager = $this->getManagerForClass('Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Phpcr\Route');
+
+        foreach ($this->routeBasePaths as $basepath) {
+            $parentPath = PathHelper::getParentPath($basepath.$requestedPath);
+
+            $parentRoute = $manager->find(null, $parentPath);
```
dbu: from a performance point of view, i suspect that it would be better to first build the list of parent path candidates and then try to find all of them in one single call. i think the overhead of preparing all parent paths is small compared to the overhead of several calls into the phpcr. the chances that we find more than one result is pretty small.

Doctrine/Phpcr/BaseSuggestionProvider.php
```diff
@@ -29,22 +30,22 @@
...
-    protected $routeBasePath;
+    protected $routeBasePaths;
 
     /**
      * @param ManagerRegistry $managerRegistry
-     * @param $routeBasePath
+     * @param array           $routeBasePath
      */
-    public function __construct(ManagerRegistry $managerRegistry, $routeBasePath)
+    public function __construct(ManagerRegistry $managerRegistry, array $routeBasePaths)
```
dbu: is this a BC break? or does php automatically cast a string into an array when needed? i know that `(array) 'bla'` does work but i am unsure about the typehint.

WouterJ: Fixed

Doctrine/Phpcr/BaseSuggestionProvider.php
WouterJ: Forgot about `findMany()`, changed the code

WouterJ:
Let's merge?

dbu:
thanks wouter!

PR #305, opened 2016-07-05
Update master from 1.3

dbu:


WouterJ:
@dbu I think it's better to just execute `git merge` locally instead of using a intermediate branch to merge both

dbu:
well, like this i get the full build matrix from travisci, just in case i mess something up.

the conflict was only on something that was actually changed in #303 , so conflicts indeed are solved with this process too.

PR #318, opened 2016-12-14
make burgov/key-value optional

dbu:


ElectricMaxxx:
mh, we had that discussion when introducing that exception in our extension class. We said just not displaying the form is a silent error no one would recognize, why is it different now?

dbu:
making it a hard dependency seems just wrong to me. in menu admin, we have a special configuration that is off by default that needs it. we only throw the exception if that config is on. we could do something similar here. and as seo is already relying on the form type, we would get rid of the problem completely in the sonata integration.

dbu:
okay, changed this a bit. if you agree with the new logic, i will adjust the tests as well. they currently expect the old behaviour

dbu:
ping @ElectricMaxxx what do you think about the code? if you agree i will adjust the tests to the new format and update the changelog.

src/DependencyInjection/CmfSeoExtension.php
```diff
@@ -52,7 +52,7 @@ public function load(array $configs, ContainerBuilder $container)
         $configuration = new Configuration();
         $config = $this->processConfiguration($configuration, $configs);
 
-        $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
+        $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
```
ElectricMaxxx: I think that current style is without any spaces next to `.`

src/DependencyInjection/Configuration.php
```diff
@@ -293,8 +293,18 @@ private function addFormSection($nodeBuilder)
                             ->scalarNode('seo_metadata')->defaultNull()->end()
                         ->end()
                     ->end()
+                    ->arrayNode('options')
+                        ->addDefaultsIfNotSet()
+                        ->children()
+                            ->enumNode('advanced')
```
ElectricMaxxx: With the info it sounds ok, but the isn't really speaking else.
What we do with that key-value stuff is arbitrary metadata, why not going in that direction?

src/Resources/config/services.xml
```diff
@@ -15,7 +15,7 @@
 
         <service id="cmf_seo.form.type.seo_metadata" class="Symfony\Cmf\Bundle\SeoBundle\Form\Type\SeoMetadataType">
             <argument>%cmf_seo.form.data_class.seo_metadata%</argument>
-            <argument>%cmf_seo.form_mode_orm%</argument>
+            <argument>%cmf_seo.form.options%</argument>
             <tag name="form.type" alias="seo_metadata" />
```
ElectricMaxxx: We don't need that tag anymore right?

ElectricMaxxx:
Activated StyleCI and fixed against master, can you rebase please?

src/DependencyInjection/CmfSeoExtension.php
dbu: hm, did not even change that actively. i wonder if phpstorm did that on save or what :-|

src/DependencyInjection/Configuration.php
dbu: you are very right. going for `generic_metadata`.

src/Resources/config/services.xml
dbu: if we would not, we could not use a constructor afaik. see http://symfony.com/doc/current/form/create_custom_field_type.html#creating-your-field-type-as-a-service

dbu:
rebased and fixed style issues. this is now ready to merge if the tests agree.

ElectricMaxxx:
@dbu can you please add some ExtensionTests for setting storage and options and ad it to the scheme file (with tests too)

src/Resources/config/services.xml
wouterj: Yes, tag is not deprecated. The `alias` attribute however can be removed.

ElectricMaxxx: Sure, i ment the alias and not the tag :-)

ElectricMaxxx:
@dbu can you do this one as second on your friday?

dbu:
fixed the existing tests, added some more and then fixed the stuff the tests reveiled. hope i got it right with the fixxmlconfig?

PR #322, opened 2017-01-01
increase licence year

ElectricMaxxx:


dbu:
thanks!
did you change the header in styleci? also, i heard that its possibly enough to have the year the project started so we could get rid of this ritual for the future :-)

ElectricMaxxx:
No idea how StyleCI handles that check.

PR #325, opened 2017-02-06
do not depend on the routing bundle for content basepath

dbu:
https://github.com/symfony-cmf/routing-bundle/pull/387 reveiled that we had an implicit dependency on the routing bundle for the phpcr-sitemap. this PR moves the setting to an seo bundle config

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -111,6 +111,7 @@ class Configuration implements ConfigurationInterface
                         ->canBeEnabled()
                         ->children()
                             ->scalarNode('manager_name')->defaultNull()->end()
+                            ->scalarNode('content_basepath')->defaultValue('/cms/content')->end()
                         ->end()
                     ->end()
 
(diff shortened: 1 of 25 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -111,6 +111,7 @@ private function addPersistenceSection(NodeBuilder $treeBuilder)
                         ->canBeEnabled()
                         ->children()
                             ->scalarNode('manager_name')->defaultNull()->end()
+                            ->scalarNode('content_basepath')->defaultValue('/cms/content')->end()
```
dbu: should we prepend this from the core bundle?

wouterj: 👍

dbu: but if we prepend, the default value here is null, right?

dbu:
https://github.com/symfony-cmf/core-bundle/pull/227

src/DependencyInjection/Configuration.php
wouterj: No, we can leave the default here as is. If CoreBundle is registered, the default will be overriden by the value that is prepended. If CoreBundle isn't available, the default value will be used.

dbu: cool, then lets merge, ok?

wouterj: Jep

PR #330, opened 2017-07-14
use trusty for hhvm

dbu:


dbu:
is this some regression with symfony 3.3?

ElectricMaxxx:
not that random, i think. Did a rebuild and the error occurs again. Do we know something about the error handling and exception controller on symfony 3.3?

there have been several [changes](https://github.com/symfony/twig-bundle/commits/3.3/Resources/views/Exception/exception.html.twig) I think we have to require 3.3 lokaly and make the test less sensitive.

dbu:
the build failure also happens in master, not related to this change

PR #333, opened 2017-09-08
fix web test

### LINKED ISSUE #331: fix build failure with symfony 3.3 / dev dependencies
the build on master started failing. i restarted slightly older builds that where green but they now fail too, so the reason is a change in a dependency.

ElectricMaxxx:
fix #331


ElectricMaxxx:
yes, master should be green than again.

dbu:
so the trailing unused parameter caused the problem? was that parameter a leftover from symfony 2?

ElectricMaxxx:
No the template has changed. For some reason my string i did not whant to be found was found.

PR #336, opened 2017-09-08
[POC] use make file

### LINKED ISSUE #334: Makefile to run tests
I would also add make files in bundles and components. Doing so we would be able to run tests separately and do not have to use the `DatabaseListener` out of the `phpunit.xml` which listens on a test suite on an ugly way. We simply should have something like:

```make
test_integration:
   init db
   run integration test suite

test_unit:
    run unit test suite

test: test_unit test_integration
``` 

I would see this as blueprint for he other bundles/components too.

ElectricMaxxx:
This would be my first try to use make files and instantiate database there instead of using the `DatabaseListener`.
What do you think?

fix #334

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "require": {
         "php": "^5.6|^7.0",
-        "symfony/framework-bundle": "^2.8|^3.0",
+        "symfony/framework-bundle": "^2.8|^3.0|^4.0",
         "sonata-project/seo-bundle": "^2.2",
         "doctrine/collections": "^1.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,8 @@
         }
     ],
     "require": {
-        "php": "^5.6|^7.0",
-        "symfony/framework-bundle": "^2.8|^3.0|^4.0",
+        "php": "^7.1",
+        "symfony/framework-bundle": "^2.8|^3.0",
         "sonata-project/seo-bundle": "^2.2",
         "doctrine/collections": "^1.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -14,11 +14,11 @@
     "require": {
         "php": "^5.6|^7.0",
         "symfony/framework-bundle": "^2.8|^3.0",
-        "sonata-project/seo-bundle": "^1.1.3|^2.0",
+        "sonata-project/seo-bundle": "^2.2.1",
         "doctrine/collections": "^1.0"
     },
     "require-dev": {
-        "symfony-cmf/testing": "^2.0",
+        "symfony-cmf/testing": "dev-testing_scripts",
         "doctrine/phpcr-odm": "^1.4.2",
         "doctrine/orm": "^2.4",
         "symfony-cmf/core-bundle": "^2.0",
@@ -51,5 +51,7 @@
         "branch-alias": {
             "dev-master": "2.1-dev"
         }
-    }
+    },
+    "minimum-stability":"dev",
+    "prefer-stable": true
 }
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require": {
         "php": "^5.6|^7.0",
         "symfony/framework-bundle": "^2.8|^3.0|^3.3|^4.0",
-        "sonata-project/seo-bundle": "^2.2.1",
+        "sonata-project/seo-bundle": "^2.2",
         "doctrine/collections": "^1.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,7 @@
         "php": "^7.1",
         "symfony/framework-bundle": "^2.8|^3.0",
         "sonata-project/seo-bundle": "^2.2",
-        "doctrine/collections": "^1.0",
-        "symfony/phpunit-bridge": "^3.3"
+        "doctrine/collections": "^1.0"
     },
     "require-dev": {
         "symfony-cmf/testing": "^2.1",
@@ -26,6 +25,7 @@
         "symfony-cmf/routing-bundle": "^1.4.0|^2.0",
         "matthiasnoback/symfony-dependency-injection-test": "^0.7.4",
         "matthiasnoback/symfony-config-test": "^1.3.1",
+        "symfony/phpunit-bridge": "^3.3.10",
         "burgov/key-value-form-bundle": "^1.0"
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "require": {
         "php": "^5.6|^7.0",
-        "symfony/framework-bundle": "^2.8|^3.0",
+        "symfony/framework-bundle": "^2.8|^3.0|^3.3|^4.0",
         "sonata-project/seo-bundle": "^2.2.1",
         "doctrine/collections": "^1.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
         "php": "^7.1",
         "symfony/framework-bundle": "^2.8|^3.0",
         "sonata-project/seo-bundle": "^2.2",
-        "doctrine/collections": "^1.0"
+        "doctrine/collections": "^1.0",
+        "symfony/phpunit-bridge": "^3.3"
     },
     "require-dev": {
         "symfony-cmf/testing": "^2.1",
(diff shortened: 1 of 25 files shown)


Makefile
```diff
@@ -0,0 +1,44 @@
...
+# For the full copyright and license information, please view the LICENSE  #
+# file that was distributed with this source code.                         #
+############################################################################
+
+CONSOLE=vendor/symfony-cmf/testing/bin/console
+
+list:
+	@echo 'test:                  will run all tests'
+	@echo 'test_unit:             will run unit tests only'
+	@echo 'test_integration_orm:  will run integration tests on PHPCR'
```
dbu: test_integration_phpcr?

Makefile
```diff
@@ -0,0 +1,44 @@
...
+	@echo 'test_integration_orm:  will run integration tests on ORM'
+
+test_integration_phpcr:
+	@echo
+	@echo '+++ create PHPCR +++'
+	${CONSOLE} doctrine:phpcr:init:dbal --drop --force
+	${${CONSOLE} doctrine:phpcr:init:dbal --drop
```
dbu: is this line needed?

tests/Functional/Doctrine/Phpcr/SeoMetadataTest.php
```diff
@@ -20,6 +21,7 @@ class SeoMetadataTest extends BaseTestCase
 {
     public function setUp()
     {
+        (new PHPCRPurger($this->getDbManager('PHPCR')->getOm()))->purge();
```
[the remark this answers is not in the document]
dbu: afaik the fixture loader itself has a flag like that already, so maybe its just a matter of passing that information to the fixture loader. or we do a functional test base class with a method to purge the database.

Makefile
```diff
@@ -0,0 +1,44 @@
...
+	${CONSOLE} doctrine:schema:create --env=orm
+	@echo '+++ run PHPCR integratin tests +++'
+	vendor/bin/simple-phpunit --testsuite "orm"
+	${CONSOLE} doctrine:database:drop --force
+
+
+test_unit:
+	@echo
+	@echo '+++ run unit tests +++'
+	vendor/bin/simple-phpunit --testsuite "Symfony SeoBundle Test Suite"
```
dbu: should we rename this suite to "unit tests" or something like that in phpunit.xml? the current name does not indicate that its only unit tests.

Makefile
ElectricMaxxx: i don't think so.

tests/Functional/Doctrine/Phpcr/SeoMetadataTest.php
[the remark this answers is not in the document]
ElectricMaxxx: i used it in the other test, where i really used the fixture loader.

Makefile
```diff
@@ -0,0 +1,50 @@
...
+	vendor/bin/simple-phpunit --testsuite "orm"
+	${CONSOLE} doctrine:database:drop --force
+
+
+test_unit:
+	@echo
+	@echo '+++ run unit tests +++'
+	vendor/bin/simple-phpunit --testsuite "unit tests"
+
+test_installation:
```
ElectricMaxxx: @dbu this would be a idea to test the installation on the branch itself.

dbu: that looks quite cool!

i wonder if we can put most of whats in this makefile into shell scripts in symfony-cmf/testing. the orm/phpcr flag could be a parameter to the script.

i would take the installation test out of the default set of tests. locally it won't work unless the commit has been pushed, and we don't need that in each combination of the matrix. we could have just one test case defined in the matrix in .travis.yml that runs this instead of the normal test suite.

ElectricMaxxx: Indeed, i will set an own matrix entry for that one. 

yes we should exclude the scripts into own scripts if they get longer than 3-4 lines. Same for travis. I forked sonata's `dev-kit` yesterday and will put that stuff into that application. So we can serve all packages with Makefiles and travis files. this one should be a blueprint to generalize it next week, when we have got Mayflower Barcamp :-)

ElectricMaxxx:
@dbu you got an idea, why the hhvm build fails now?

Makefile
dbu: why not put the scripts into symfony-cmf/testing? they are about testing, after all... afaik the dev-kit is a way to keep boilerplate files of each repo in sync. that is the next best option, but i think using symfony-cmf/testing would be simpler and work just as well.

dbu:
> hhvm

had no idea either, so i re-started the last successful test of the master branch. it fails with the same error, so it looks like a setup change at travis-ci rather than something of this branch. we said we keep hhvm support as long as the hassle is not too big - sounds to me like the moment to drop it, honestly.

Makefile
ElectricMaxxx: Sure. But i can do one step after each only. So:

1. Getting this working in complete
2. Move tasks to scripts in `testing`
3. Serve Makefile and .travis.yaml from a dev-kit, calling scripts in `testing` to all of our bundles?
4. Fix broken stuff
5. Happy

I would even move the `testing-install` stuff into `testing` later. Currently it is the easiest way to get an overview an serve changes.

ElectricMaxxx:
btw: the broken installation build depends on the release of: https://github.com/sonata-project/SonataSeoBundle/commit/181215ef16f29eded95cc79d5a2cb7f4c7ca1316

Makefile
dbu: sounds like a great plan!

ElectricMaxxx:
I am now behind `master-dev-kit` with this, as dev-kit introduces the makefile and tranvis checks there. I also added Symfony 3.4 and 4.0 there lets see what happens :-)

ElectricMaxxx:
aaaand ... we see the stricter style checks now.

ElectricMaxxx:
argh, now we collide with sonata's php 7.1 requirement. We have to follow now too.

ElectricMaxxx:
Except the fail on prefer lowest and symfony 4.0 it looks good with php 7.1. Both should be caused by sonata-seo. I would like to get 2.2.1, which isn't released jet. But dumping php 7.0 and 5.6 sounds like a BC for me.

dbu:
> But dumping php 7.0 and 5.6 sounds like a BC for me.

its not. composer will prevent from installing on outdated systems. lets jump to php 7.1. people on legacy system can keep using the 2.0 version of seo and sonata.

ElectricMaxxx:
i also added `symfony: ^4.0, and also did on testing component. But it fails caused by other dependencies, should i keep it or is `^3.4` enough?

ElectricMaxxx:
btw: @dbu you know a fix for our red prefer-lowest build:
https://travis-ci.org/symfony-cmf/seo-bundle/jobs/289844120
some other phpunit version?

dbu:
lets get things working as is and add symfony 4 later in a separate pull request to not complicate things.

about the build failure: thats really odd. the phpunit bridge is installing phpunit and we get the exact same version. however, something also installs sebastian/exporter as a composer dependency. can you check with `composer why` which dependency causes that? i would have thought that phpunit would use its own classes also for dependencies, rather than those of the project, but that seems not to work that way.

tests/Functional/Doctrine/Phpcr/SeoMetadataTest.php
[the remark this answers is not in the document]
ElectricMaxxx: found that the loader purges every time, the parameter decides if the initializer is called.

ElectricMaxxx:
Awesome, moving `assertEquals` to `assertSame` breaks some tests.

composer.json
```diff
@@ -15,7 +15,8 @@
         "php": "^7.1",
         "symfony/framework-bundle": "^2.8|^3.0",
         "sonata-project/seo-bundle": "^2.2",
-        "doctrine/collections": "^1.0"
+        "doctrine/collections": "^1.0",
+        "symfony/phpunit-bridge": "^3.3"
```
dbu: this should stay in require-dev. for installing this as a dependency, we must not depend on the phpunit-bridge

dbu:
this is better, but not fully good. can you try `^3.3.10` to get the latest 3.3 version?

dbu:
still same problem. can you try bumping the noback things to the latest version?

ElectricMaxxx:
I think our tests are not compatible with phpunit 6.*, or?

dbu:
i guess not. if all php versions we want to test with are compatible with phpunit 6, we should just upgrade to that. but if not, lets find something that is new enough to work but not too new...

ElectricMaxxx:
so i will try to go on phpunit 6.0, cause there is no version of the mathiasnoback stuff in between.

ElectricMaxxx:
i will merge this one into the `master-dev-kit` to have one open PR here only.

PR #341, opened 2017-10-27
Master dev kit

ElectricMaxxx:
This one should fix the form type test.


ElectricMaxxx:
To fix the last deprecation, https://github.com/doctrine/DoctrinePHPCRBundle/issues/285 should be done.

tests/Unit/Form/Type/SeoMetadataTypeTest.php
```diff
@@ -14,24 +14,42 @@
...
  * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
  */
-class SeoMetadataTypeTest extends TypeTestCase
+class SeoMetadataTypeTest extends FormIntegrationTestCase
 {
+    private $validator;
+
+    public function setUp()
+    {
+        $this->validator = $this->createMock(ValidatorInterface::class);
```
ElectricMaxxx: we have to set it up on our own, as the base class in gone (moved to a trait) in symfony 3.4

src/Controller/SuggestionProviderController.php
```diff
@@ -116,15 +119,6 @@ public function listAction(
     }
 
     /**
-     * @param SuggestionProviderInterface $matcher
-     * @param string                      $group
-     */
-    public function addSuggestionProvider(SuggestionProviderInterface $matcher, $group)
```
dbu: this is a BC break if somebody relied on adding providers later, e.g. in a compilar pass

src/DependencyInjection/CmfSeoExtension.php
```diff
@@ -34,11 +34,6 @@ class CmfSeoExtension extends Extension
      */
     private $defaultAlternateLocaleProviderId;
 
-    /**
-     * @var bool Whether the content listener is loaded
-     */
-    private $contentListenerEnabled = false;
```
dbu: was this just unused?

ElectricMaxxx: yes, and sensio quality guard mentioned it.

src/Controller/SuggestionProviderController.php
ElectricMaxxx: mhh, got a hint from sensio guard not to use other public methods then actions in controllers, what to do? Changelog enough?

ElectricMaxxx: ahh you wrote it :-)

ElectricMaxxx: Added a BC-BREAK notice in change log, i think we can live ẃith that on an non major release.

PR #349, opened 2018-06-30
DevKit updates - Prepare

ElectricMaxxx:
Prepare to release and use SF best practice on travis.


ElectricMaxxx:
I should not mix that suff up, i will do release first and then introduce symfony best practices on travis.yml

